PR #1, opened 2016-03-11
Made script for auto-upload

JureZmrzlikar:


mstajdohar:
Also, change the [_upload_file](https://github.com/genialis/resolwe-bio-py/blob/master/resolwe_api/resolwe.py#L299) method to be compliant with the [new upload protocol](https://github.com/genialis/genialis-base/pull/18/files#diff-ee7f6e6d45b6b72952f6243f37daefb0R108) introduced by @kostko.

mstajdohar:
Read about [PEP 8](https://www.python.org/dev/peps/pep-0008/) and [PEP 257](https://www.python.org/dev/peps/pep-0257/) and run the script through [pep8](https://pypi.python.org/pypi/pep8) and [PyLint](https://pypi.python.org/pypi/pylint) with configuration files [.pep8rc](https://github.com/genialis/resolwe/blob/master/.pep8rc) and [.pylintrc](https://github.com/genialis/resolwe/blob/master/.pylintrc).

scripts/sequp.py
```diff
@@ -0,0 +1,282 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import os
+from os.path import expanduser
+import time
+import fnmatch
+import argparse
+import csv
+
+import appdirs
```
mstajdohar: Not listed in install_requires in `setup.py`.

JureZmrzlikar:
added appdirs in install_requires in setup.py

mstajdohar:
@dblenkus, rebase and fix "just temp commit" before accepting this. Please make it compatible with the new version of Resolwe. I can work on the additional annotation fields tomorrow.

PR #2, opened 2016-04-13
Full annotation support for upload from sequencer

mstajdohar:
Requires [resolwe PR #56](https://github.com/genialis/resolwe/pull/56).


mstajdohar:
Works OK, but not yet bulletproof. @JureZmrzlikar, extensive tests will be needed.

PR #3, opened 2016-04-14
Upload gzipped files

mstajdohar:


resolwe_api/resolwe.py
```diff
@@ -1,5 +1,5 @@
 """Resolwe"""
-from __future__ import absolute_import, division, print_function, unicode_literals
```
kostko: Did you find out what exactly causes the crash in case `unicode_literals` are used?

kostko:
LGTM

resolwe_api/resolwe.py
dblenkus: Does this also work in py3?

mstajdohar: @kostko: no, @dblenkus: yes

PR #4, opened 2016-04-29
Write tests for resolwe.py and make corresponding API changes.

JureZmrzlikar:


resolwe_api/resolwe.py
```diff
@@ -14,7 +14,9 @@
     from urllib import parse as urlparse
 
 import requests
+from requests.exceptions import ConnectionError
```
mstajdohar: Use one or the other option. If you really wish to use it bot ways, group import and from sentences separately. Read about structuring imports [here](https://www.python.org/dev/peps/pep-0008/#imports).

resolwe_api/resolwe.py
```diff
@@ -61,6 +57,9 @@ def collections(self):
             self.cache['collections'] = {c['id']: Collection(c, self) for c in self.api.collection.get()}
         return self.cache['collections']
 
+        # TODO what if person does not have access to some data objects?
```
mstajdohar: Good point another argument in favor of removing the cache implementation.

resolwe_api/resolwe.py
```diff
@@ -35,21 +37,15 @@ def __init__(self, email=DEFAULT_EMAIL, password=DEFAULT_PASSWD, url=DEFAULT_URL
...
-        # collections = dict of (collection_id - collection) pairs
-        # collections_data - dict of (collection_id - data_list) pairs,
-        #   where data_list is list of all data objects for given collection_id
+        # data = dict of (data_id - Data object) pairs
+        # collections = dict of (collection_id - Collection object) pairs
+        # collections_data - dict of (collection_id - Data objects list) pairs,
         self.cache = {'data': {}, 'collections': None, 'collections_data': {}}
-        # How about when the sever is changed - how can we get the changes? For example someone adds a collection or some data. Cache does not know this and cannot update?
-        # this cache thing is just partially implemented now...
+        # TODO: what if there are updates on server? How to know when to update cache?
```
mstajdohar: The caching will have to be rethought. It was designed to support caching of the whole collection locally. We expect to access samples directly in the future. Maybe we should remove the cache altogether and reintroduce it back later when we will have a better idea of the new workflows.

resolwe_api/resolwe.py
```diff
@@ -189,16 +213,18 @@ def processes(self, process_name=None):
...
         else:
             return self.api.process.get()
 
-    def printad_processes(self):
+    def print_upload_processes(self):
         """Print all upload process names."""
         for p in self.processes():
             if 'upload' in p['category']:
-                print(p['name'])
+                sys.stdout.write(p['name'] + '\n')
```
mstajdohar: Why is this better?

resolwe_api/resolwe.py
```diff
@@ -211,18 +237,24 @@ def print_process_inputs(self, process_name):
...
+            raise Exception('Invalid process name: {}'.format(process_name))
 
         for field_schema, _, _ in iterate_schema({}, p['input_schema'], 'input'):
             name = field_schema['name']
             typ = field_schema['type']
-            # value = fields[name] if name in fields else None
-            print("{} -> {}".format(name, typ))
+            sys.stdout.write("{} -> {}".format(name, typ))
 
     def rundata(self, strjson):
```
mstajdohar: Remove this function. We will probably want a similar function in the future, _e.g._ `run(process_name, inputs)` that will return a data object. This would be similar to the `create` function, but more explicit that it runs the process in the backend. Anyway, let's remove it for now and wait for the bioinformatician's workflows.

resolwe_api/resolwe.py
```diff
@@ -245,19 +277,11 @@ def create(self, data, resource='data'):
             data = json.dumps(data)
 
         if not isinstance(data, str):
```
mstajdohar: Does this work with Python3? Write a test. In general, we should use `six.string_types` instead of `str` to support Python3.

resolwe_api/resolwe.py
```diff
@@ -283,32 +307,36 @@ def upload(self, process_name, name='', descriptor=None, descriptor_schema=None,
...
-            Exception('Invalid process name {}'.format(process_name))
-
-        for field_name, field_val in fields.items():
-            if field_name not in p['input_schema']:
-                Exception("Field {} not in process {} inputs".format(field_name, p['name']))
-
-            if find_field(p['input_schema'], field_name)['type'].startswith('basic:file:'):
-                if not os.path.isfile(field_val):
-                    Exception("File {} not found".format(field_val))
+            raise Exception('Invalid process name: {}'.format(process_name))
```
mstajdohar: Exception messages sometimes start with a capital letter (like here) and sometimes with a lower case letter (in the `raise ValueError` above). Find out the suggested way and be consistent.

resolwe_api/resolwe.py
```diff
@@ -283,32 +307,36 @@ def upload(self, process_name, name='', descriptor=None, descriptor_schema=None,
...
 
         inputs = {}
 
         for field_name, field_val in fields.items():
-            if find_field(p['input_schema'], field_name)['type'].startswith('basic:file:'):
+            input_fields = dict([(e['name'],e['type']) for e in p['input_schema']])
```
mstajdohar: Make code style compliant with PEP 8.

resolwe_api/resolwe.py
```diff
@@ -360,7 +388,7 @@ def _upload_file(self, fn):
 
                 for i in range(5):
                     if i > 0 and response is not None:
-                        print("Chunk upload failed (error {}): repeating for chunk number {} ."
+                        sys.stdout.write("Chunk upload failed (error {}): repeating for chunk number {} ."
```
mstajdohar: Why is that better than print?

resolwe_api/resolwe.py
```diff
@@ -390,14 +418,15 @@ def _upload_file(self, fn):
                 sys.stdout.write("\n{:.0f} % Uploaded {}".format(progress, fn))
                 sys.stdout.flush()
                 chunk_number += 1
-        print()
+
+        sys.stdout.write("\n")
```
mstajdohar: Print?

resolwe_api/resolwe.py
```diff
@@ -407,52 +436,50 @@ def download(self, data_objects, field):
...
 
         try:
             response = requests.post(urlparse.urljoin(url, '/rest-auth/login/'), data=payload)
-        except requests.exceptions.ConnectionError:
+        except ConnectionError:
             raise Exception('Server not accessible on {}'.format(url))
 
-        if response.status_code == 403:
+        if response.status_code in [400, 403]:
             raise Exception('Invalid credentials.')
```
mstajdohar: Use the following error message: Response HTTP status code {}. Invalid credentials?

resolwe_api/resolwe.py
```diff
@@ -407,52 +436,50 @@ def download(self, data_objects, field):
...
 
         if not ('sessionid' in response.cookies and 'csrftoken' in response.cookies):
-            raise Exception('Invalid credentials.')
+            raise Exception('Missing sessionid / csrftoken.')
```
mstajdohar: Use the following error message: Missing sessionid or csrftoken. Invalid credentials?

resolwe_api/tests/test_resolwe.py
```diff
@@ -0,0 +1,372 @@
+import os
+import unittest
+
+from mock import patch, MagicMock
+import requests
+from requests.exceptions import HTTPError
+from slumber.exceptions import HttpNotFoundError
+
```
mstajdohar: Fix code style to be compliant with PEP 8.

mstajdohar:
We should automate unit testing and style checks on [travis-ci.org](travis-ci.org) that is free for open source projects. Add `.travis.yml`, `tox.ini`?, `.pylintrc`, `.pep8rc` confiration files (see examples in [docker-bio-linux8-resolwe](https://github.com/genialis/docker-bio-linux8-resolwe) and [resolwe](https://github.com/genialis/resolwe)). Ask @tjanez and @dblenkus for help.

mstajdohar:
Fix usage examples in README.rst, they still import `Genesis`.

mstajdohar:
When this PR is accepted, @dblenkus should bump the version and publish on PyPI. Also, we should then add badges to the top of the `README.rst` file [like here](https://github.com/genialis/resolwe).

resolwe_api/resolwe.py
JureZmrzlikar: I thought that it is not possible to mock print function. When I was running tests, I ended up with a bunch of print statements in the terminal... Only now I see that it is possible to mock print. Should I change all stdout to print?

resolwe_api/resolwe.py
JureZmrzlikar: Completely removing the cache means a huge change. Should we still do it, or keep it this way until we know what we want? I like the second idea better :-)

resolwe_api/resolwe.py
mstajdohar: Maybe change it to [Python's logger](https://docs.python.org/2/library/logging.html). We will need to log the output when it runs as a daemon on the server. It would be nice if you could tell it with some setting where to log. The default could be stdout. @dblenkus, what do you think about it?

resolwe_api/resolwe.py
mstajdohar: I will remove it in a separate PR. Huge change indeed, but it will reduce the amount of code significantly and it will make it easier for us to maintain the code quality and easier to test for correctness. Caching introduces many problems, so we should carefully think how to support it.

mstajdohar: It would be great if we could use a 3rd-party library for caching, like [requests-cache](https://pypi.python.org/pypi/requests-cache). This would really reduce the complexity of our code and lower the chance of error.

resolwe_api/resolwe.py
JureZmrzlikar: Actually you always have to raise them. The deleted code could not raise the exceptions at all.

mstajdohar: The comment is about messages. Here you write `'Invalid process name: {}'`, but maybe you should write `'invalid process name: {}` to be consistent with the lowercase message in line 280.

JureZmrzlikar:
I addressed the following comments:
    * imports formatting
    * remove rundata function
    * make code and tests compatible with Python 3
    * format exception messages
    * make code pep8 compliant
    * Fix usage examples in README.rst

Issues remaining:
    * remove cache
    * port sys.stdout to logger
    * automate unit testing and style checks on travis-ci.org
    * dblenkus should bump the version and publish on PyPI
    * add badges to the top of the README.rst

I suggest them to be solved in separate PR.

PR #6, opened 2016-05-05
Find and download BAM.

JureZmrzlikar:
- Remove functions cllections(), collection_data() and data() from Resolwe class 
- Add three namespaces (data, sample, collection) to Resolwe class
- Make ResolweQuerry class and implement get & find methods
- Make Sample class and implement methods files() and download()
- Write some documentation for new added functions


resolwe_api/resolwe.py
```diff
@@ -493,3 +337,102 @@ def __call__(self, request):
...
+    ResolweQuerry class. ResolweQuerry enables querries on
+    corresponding objects, for example:
+
+    re.data.get(42) # return Data object with ID 42.
+    re.sample.filter(contributor=1) # return all samples made by contributor 1
+
+    Detailed description of methods can be found in method docstrings.
+    """
+
+    def __init__(self, resolwe, type_):
```
mstajdohar: I would call `type_` a `Resource`. This would be aligned with [REST](https://en.wikipedia.org/wiki/Representational_state_transfer) philosophy.

resolwe_api/resolwe.py
```diff
@@ -493,3 +337,102 @@ def __call__(self, request):
...
+        self.resolwe = resolwe
+        self.object = type_
+        self.type = type_.endpoint
```
mstajdohar: I would just save this as `self.endpoint`.

resolwe_api/resolwe.py
```diff
@@ -493,3 +337,102 @@ def __call__(self, request):
...
+        # Make separate sluber API instance, since the one from
+        # Resolwe colud not be used as desired.
+        self.connection = slumber.API(self.resolwe.url,
```
mstajdohar: Are you sure you need to create another connection? I am quite sure `self.resolwe.api` should work just fine.

resolwe_api/resolwe.py
```diff
@@ -493,3 +337,102 @@ def __call__(self, request):
...
+                                      self.resolwe.auth,
+                                      append_slash=False,
+                                      )
+
+    def get(self, uid):
+        """
+        Get object with provided ID or slug.
+
+        :param uid: unique identifier - ID or slug
+        :type process_name: int(ID) or string(slug)
```
mstajdohar: Rename to `:type uid:`

resolwe_api/resolwe.py
```diff
@@ -493,3 +337,102 @@ def __call__(self, request):
...
+
+    re.data.get(42) # return Data object with ID 42.
+    re.sample.filter(contributor=1) # return all samples made by contributor 1
+
+    Detailed description of methods can be found in method docstrings.
+    """
+
+    def __init__(self, resolwe, type_):
+        self.resolwe = resolwe
+        self.object = type_
```
mstajdohar: I would save this as `self.Resource`.

mstajdohar:
@JureZmrzlikar, wait for @kostko's feedback before you address my comments.

resolwe_api/sample.py
```diff
@@ -0,0 +1,138 @@
+"""Sample"""
```
mstajdohar: Maybe we should have a module called `resources` that would contain `data.py`, `sample.py`, `collection.py`, and later `study.py`.

resolwe_api/sample.py
```diff
@@ -0,0 +1,138 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import os
+import six
+
+DOWNLOAD_TYPES = {
```
mstajdohar: The `download()` method should also be implemented on Data and Collection. I would move this dict to `resources/base.py` and import it in `data.py`, `sample.py` and `collection.py`. I would think about a `BaseResource` class.

resolwe_api/sample.py
```diff
@@ -0,0 +1,138 @@
...
+
+        elif name:
+            dfiles = [d for d in dfiles if d[1] == name]
+
+        elif typ:
+            data_type, output_field = None, None
+            if isinstance(typ, six.string_types):
+                # typ is abbreviation (string)
+                data_type, output_field = DOWNLOAD_TYPES[typ]
+            else:
```
mstajdohar:
```
elif isinstance(typ, tuple) and len(typ) == 2:
    ...
else:
    raise ValueError('Invalid typ parameter')

File: resolwe_api/sample.py [hunk 9]
```diff
@@ -0,0 +1,138 @@
...
+
+        if not download_dir:
+            download_dir = os.getcwd()
+
+        # list of all files to download - this list is yet to be
+        # filtered by name/typ:
+        dfiles = self.files(verbose=True)
+
+        if not (name or typ) or force:
+            pass
```
mstajdohar: I guess WIP?

JureZmrzlikar: No. If user executes:
s.download(name='blah', force=True)
it would still download all files. I thought that force should have precedence over other params.

resolwe_api/resolwe.py
```diff
@@ -250,7 +94,7 @@ def print_process_inputs(self, process_name):
         for field_schema, _, _ in iterate_schema({}, p['input_schema'], 'input'):
             name = field_schema['name']
             typ = field_schema['type']
-            sys.stdout.write("{} -> {}".format(name, typ))
+            sys.stdout.write("{} -> {}".format(name, typ) + "\n")
```
kostko: Why `+ "\n"` instead of just modifying the original string by adding a `\n`, so `"{} -> {}\n".format(name, typ)`.

resolwe_api/resolwe.py
kostko: Why `type_`? It is not a reserved word in Python, so it may be overriden in the local scope.

resolwe_api/resolwe.py
kostko: Perhaps `self.resource`, but really not capitalized attribute names.

resolwe_api/sample.py
kostko: I agree.

resolwe_api/resolwe.py
dblenkus: Idea with capitalized attributes is from `boto3` (SKD for AWS). `Resource` is class on server, so they try to mimic that: https://boto3.readthedocs.io/en/latest/guide/quickstart.html#using-boto-3

    # Upload a new file
    data = open('test.jpg', 'rb')
    s3.Bucket('my-bucket').put_object(Key='test.jpg', Body=data)

resolwe_api/resolwe.py
JureZmrzlikar: Thanks, will fix that.

resolwe_api/resolwe.py
JureZmrzlikar: Only option I see to reuse connection is to make somthing like this:

def __init__(self, resolwe, type):
    ...
    self.api = None
    if type == Data:
        self.api = resolwe.api.data
    if type == Sample:
        self.api = resolwe.api.sample
    if type == Collection:
        self.api = resolwe.api.collection
    ...

Which i thought was ugly. But I guess making another connection is far more uglier. :-)
And since we don't expect to have more types of **type** parameter (only study at sight) it seems like an OK  solution. Any other idea?

mstajdohar: What about [getattr](https://docs.python.org/2/library/functions.html#getattr)?
```
self.api = getattr(resolwe.api, 'data')
```

resolwe_api/resolwe.py
mstajdohar: @kostko, even if they represent a class?

resolwe_api/sample.py
JureZmrzlikar: Should this be done now or in a separate PR?

mstajdohar: Move it now, but we can support the `download` of Data and Collection later.

resolwe_api/resolwe.py
kostko: Yes. I don't think that attributes should be named based on their type. And a capital letter for attribute or keyword argument name (as shown above) just seems strange to me.

PR #12, opened 2016-05-12
Develop processes locally

mstajdohar:
* Rename upload method to run and refactor to run any process
* Move downlad method from resolwe.py to resource/base.py
* Fix renaming leftovers
* Read process from YAML and post to server
* SCP auxiliary process' scripts (tools) to server

### DIFF
--- a/resdk/resolwe.py
+++ b/resdk/resolwe.py
@@ -108,6 +108,7 @@ class Resolwe(object):
         :type collections: list of ints
         :param data_name: Default name of Data object
         :type data_name: string
+
         :rtype: HTTP Response object
 
         """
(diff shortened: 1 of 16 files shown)


mstajdohar:
@JureZmrzlikar, I would kindly ask you to add tests for new methods: run (in resolwe.py) and download (in resources/base.py). Look back in commit history on Github if you need examples of previous implementations of upload and download tests.

resdk/resolwe.py
```diff
@@ -92,115 +84,88 @@ def print_process_inputs(self, process_name):
...
+        if ((descriptor and not descriptor_schema) or
+                (not descriptor and descriptor_schema)):
+            raise ValueError("Set both or neither descriptor and descriptor_schema")
 
-        process = self.processes(process_name=process_name)
+        process = self.api.process.get(slug=slug, ordering='-version', limit=1)
 
         if len(process) == 1:
             process = process[0]
-        elif len(process) > 1:
```
JureZmrzlikar: This block of code handles the situation when there are two versions of one process (two equal slugs for two different versions of same process). It just ensures that the latest version of processor is taken. Multiple versions of processes exist becouse of backward compatibility (on genesis), at least i was told so by @dblenkus.
Anyway, I think this part should be kept.

PR #14, opened 2016-05-16
Test run and download methods

JureZmrzlikar:


resdk/resources/utils.py
```diff
@@ -11,7 +11,7 @@ def iterate_fields(fields, schema):
 
     """
     schema_dict = {val['name']: val for val in schema}
-    for field_id, properties in fields.iteritems():
+    for field_id, properties in fields.items():
```
mstajdohar: I see that tests have yield the first results. Good job!

mstajdohar:
@JureZmrzlikar, I think this work on tests is very important. Nevertheless it will wait until the end of current milestone. It has to be reviewed well, but Domen and I are busy building the requested features.

mstajdohar:
@JureZmrzlikar, could split the two bug fixes to a separate commit and create a separate PR? I wish to accept those tomorrow.

dblenkus:
@JureZmrzlikar I think that tests are too long. Try to keep them as short as possible (few lines, at most one screen) and put common logic in `setUp` function. This way they will be more transparent, quicker to run (at least one test) and easier to isolate bug.

mstajdohar:
Good point @dblenkus.

PR #26, opened 2016-06-09
Unit tests for Resolwe and ResAuth class

JureZmrzlikar:


resdk/tests/unit/test_resolwequerry.py
```diff
@@ -4,19 +4,25 @@
...
-if six.PY2:
-    # pylint: disable=deprecated-method
-    unittest.TestCase.assertRegex = unittest.TestCase.assertRegexpMatches
-
 
 class TestResolweQuerry(unittest.TestCase):
 
+    @patch('resdk.resolwe.getattr')
+    @patch('resdk.resolwe.Resolwe')
+    @patch('resdk.resources.sample.Sample')
```
mstajdohar: Do we really wish to patch everything? I would not patch parts of our SDK, but calls to external functions (if necessary). If you would not patch Sample here, you would cover some part of the Sample class as well. Is this bad? @dblenkus, what do you think?

JureZmrzlikar: In general, i understand that unit tests should cover only the function under test, not the whole package interaction... However, I agree that some interconnection can probably be good for bug recognition.

In this particular case, I just wanted a mock with "endpoint" attribute. I could created it manually, but using a real resource just seemed more explicit.

dblenkus: Yes, unit test should mock everything that comes from outside. It is good approach to write few e2e tests that cover common use cases and then unit tests for every possibility. But I agree that sometimes it is hard (or doesn't make any sense) to mock everything. So use common sense to decide what to mock and what not.

mstajdohar:
Rebase to the latest master. We have changed how the tests are called.

mstajdohar:
@dblenkus, please review.

JureZmrzlikar:
Rebased to latest master.

resdk/resources/utils.py
```diff
@@ -80,3 +80,16 @@ def _print_input_line(element_list, level):
...
                 fill_spaces("[" + element['type'] + "]", max_type_len + spacing),
                 element['label']))
+
+
+def endswith_colon(schema, field):
+    """
+    Ensure the field ends with colon.
+
+    :param schema: Register process from source YAML file
+    :type schema: type
```
dblenkus: `type` -> `dict`

resdk/tests/unit/test_resolwe.py
```diff
@@ -4,82 +4,406 @@
...
+    def test_raise_if_bad_yaml_file(self):
+        with self.assertRaises(yaml.parser.ParserError):
+            Resolwe._register(self.resolwe_mock, self.random_file, "alignment-bowtie")
+
+    def test_raise_if_slug_not_in_yaml(self):
+        message = r"Process source given .* but process slug not found: .*"
+        with six.assertRaisesRegex(self, ValueError, message):
+            Resolwe._register(self.resolwe_mock, self.yaml_file, "bad-slug")
+
+    def test_123(self):
```
dblenkus: Add docstring with short test description, if test name is not descriptive.

resdk/resolwe.py
```diff
@@ -155,29 +147,29 @@ def endswith_colon(schema, field):
...
+                        self._version_int_to_string(process['version']))
+
+                response = self.api.process.post(process)
 
             elif len(server_process) == 0:
                 response = self.api.process.post(process)
             else:
                 raise ValueError("Unexpected behaviour at get process with slug {}".format(slug))
 
+        # How to reproduce this? What scenario causes this exception?
```
dblenkus: Updating processes is supported only on development servers, so this error will be raised on production server.

dblenkus:
Add some comments/docstrings to tests and structure them a bit, so it will be more obvious what they are testing. Especially tests with a function call or two and then a lot of asserts are confusing.

JureZmrzlikar:
Added some comments/docstrings and some more structure. Hope it looks better now.

mstajdohar:
@dblenkus, please have another review pass.

JureZmrzlikar:
This PR is now modified and contains 6 new commits - quite some new content...

PR #33, opened 2016-06-28
Fix resdk

JureZmrzlikar:


jkokosar:
@JureZmrzlikar How does omitting "output" influence the possible download of files from the data object _input_ fields? Was that ever even intended functionality - to also enable download of files from the input fields?

JureZmrzlikar:
Currently (as far as i know) there is no possibility to download inputs. If one wishes to do so, it is best to check the corresponding sample (sample includes also the parent data object) and download from there. I do not know if we envision to overcome this in the future.

resdk/resolwe.py
```diff
@@ -508,9 +509,15 @@ def get(self, uid):
 
         """
         if re.match('^[0-9]+$', str(uid)):  # iud is ID number:
```
dblenkus: Use `isdigit` method: https://docs.python.org/3/library/stdtypes.html#str.isdigit

resdk/resolwe.py
```diff
@@ -508,9 +509,15 @@ def get(self, uid):
...
+            if self.endpoint == 'presample':
+                return self.resource(id=uid, resolwe=self.resolwe, presample=True)
+            else:
+                return self.resource(id=uid, resolwe=self.resolwe)
         else:  # uid is slug
-            return self.resource(slug=uid, resolwe=self.resolwe)
+            if self.endpoint == 'presample':
+                return self.resource(slug=uid, resolwe=self.resolwe, presample=True)
+            else:
+                return self.resource(slug=uid, resolwe=self.resolwe)
```
dblenkus: Write this like:

    sample_filter = {'id': uid} if uid.isdigit() else {'slug': uid}
    if self.endpoint == 'presample':
        sample_filter['presample'] = True
    return self.resource(resolwe=self.resolwe, **sample_filter)

dblenkus:
@jkokosar if inputs are files, they are not actually stored on server. They must be processed by process and stored in `output` if you want to access them later. If inputs are Data objects, as @JureZmrzlikar pointed out, you can access that object and downloads it's outputs. 
If this is a common use-case (that someone needs outputs of an input object), you should present it and discuss it with @mstajdohar, so we can place it properly in our development cycle, so they don't interfere other tasks.

dblenkus:
LGTM

jkokosar:
@dblenkus I was thinking about Data objects, so the workflow @JureZmrzlikar and you proposed is sufficient.

PR #36, opened 2016-07-27
Fix resolwe.run docstring

JureZmrzlikar:


mstajdohar:
Nice work, the code is great. Please add tests.

JureZmrzlikar:
This PR only changes one docstring and two pylint errors. I don't think that tests should be required for this. But yes - the other PR will significantly increase coverage, as soon as e2e testsing is supported.

mstajdohar:
Tests should be added because they were not written in the first place (for the three lines that you changed). Are this lines covered in e2e tests? If not, we should write a unit test for this.

JureZmrzlikar:
Yes, they are covered in e2e tests.

dblenkus:
The code with missing parenthesis shouldn't work. Does e2e test catch this?

JureZmrzlikar:
Tests cover changed lines, so i believe they would fail.

mstajdohar:
If we have a test for this in e2e it LGTM.

PR #40, opened 2016-08-17
Fix linters

mstajdohar:
Check PEP 8 and PEP 257. Fix lint errors and warnings.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -5,6 +5,15 @@ Change Log
 All notable changes to this project are documented in this file.
 
 
+==========
+Unreleased
+==========
+
+Added
+-----
+* Check PEP 8 and PEP 257
+
+
 ==================
 1.3.6 - 2016-08-15
 ==================
--- a/resdk/resolwe.py
+++ b/resdk/resolwe.py
@@ -1,4 +1,5 @@
-"""
+""".. Ignore pydocstyle D400.
+
 =======
 Resolwe
 =======
@@ -20,10 +21,10 @@ import subprocess
 
 import yaml
 import requests
+# Needed because we mock requests in test_resolwe.py
+from requests.exceptions import ConnectionError  # pylint: disable=redefined-builtin
 import slumber
-
 from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
-from requests.exceptions import ConnectionError  # pylint: disable=ungrouped-imports, redefined-builtin
 
 from .resources import Data, Collection, Sample, Process
 from .resources.utils import iterate_fields, iterate_schema, endswith_colon
@@ -40,7 +41,6 @@ TOOLS_REMOTE_HOST = os.environ.get('TOOLS_REMOTE_HOST', None)
 
 
 class Resolwe(object):
-
     """Connect to a Resolwe server.
 
     :param email: user's email
@@ -53,6 +53,7 @@ class Resolwe(object):
     """
 
     def __init__(self, email=DEFAULT_EMAIL, password=DEFAULT_PASSWD, url=DEFAULT_URL):
+        """Initialize attributes."""
         self.url = url
         self.auth = ResAuth(email, password, url)
         self.api = slumber.API(urljoin(url, '/api/'), self.auth, append_slash=False)
@@ -70,7 +71,7 @@ class Resolwe(object):
         version_numbers = [int(number_string) for number_string in version.split(".")]
 
         if len(version_numbers) > len(VERSION_NUMBER_BITS):
-            raise NotImplementedError("Versions with more than {0} decimal places are not supported.".format(
+            raise NotImplementedError("Version should have at most {} decimal places.".format(
                 len(VERSION_NUMBER_BITS) - 1))
 
         # add 0s for missing numbers
@@ -82,7 +83,7 @@ class Resolwe(object):
         for num, bits in zip(reversed(version_numbers), reversed(VERSION_NUMBER_BITS)):
             max_num = (bits + 1) - 1
             if num >= 1 << max_num:
-                raise ValueError("Number {0} cannot be stored with only {1} bits. Max is {2}.".format(
+                raise ValueError("Number {} cannot be stored with only {} bits. Max is {}.".format(
                     num, bits, max_num))
             version_number += num << total_bits
             total_bits += bits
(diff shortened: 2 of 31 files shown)


mstajdohar:
The lines that decrease the test coverage are part of the e2e tests that are not yet enabled on Travis.

mstajdohar:
@tjanez, linters fixed, have a look.

resdk/tests/unit/test_collections.py
```diff
@@ -29,7 +29,8 @@ def test_init(self, collection_mock):
 
     @patch('resdk.resources.collection.BaseCollection', spec=True)
     def test_data_types(self, collection_mock):
-        api_mock = MagicMock(**{'data.return_value': MagicMock(**{'get.return_value': DATA_SAMPLE[0]})})
+        api_mock = MagicMock(**{'get.return_value': DATA_SAMPLE[0]})
```
tjanez: I think using `api_mock` variable name twice is unnecessarily confusing. Can you change one name to something else?

resdk/tests/unit/test_resolwe.py
```diff
@@ -68,8 +68,9 @@ class TestRegister(unittest.TestCase):
 
     @patch('resdk.resolwe.Resolwe', spec=True)
     def setUp(self, resolwe_mock):  # pylint: disable=arguments-differ
-        self.yaml_file = os.path.normpath(os.path.dirname(__file__) + "./../files/bowtie.yaml")
-        self.bad_yaml_file = os.path.normpath(os.path.dirname(__file__) + "./../files/bowtie_bad.yaml")
+        base_dir = os.path.dirname(__file__)
```
tjanez: I think it would be even better to actually save the common base directory to the `base_dir` variable:
```python
base_dir = os.path.abspath(os.path.dirname(__file__) + '/..')
```

and then just use:
```python
self.yaml_file = os.path.join(base_dir, 'files', 'bowtie.yaml')
self.bad_yaml_file = os.path.join(base_dir, 'files', 'bowtie_bad.yaml')
```

resdk/tests/unit/test_resolwe.py
```diff
@@ -326,7 +331,7 @@ def test_call_with_all_args(self, resolwe_mock, os_mock, data_mock):
 class TestUploadFile(unittest.TestCase):
 
     def setUp(self):
-        self.fn = os.path.normpath(os.path.dirname(__file__) + "./../files/example.fastq")
+        self.file_path = os.path.normpath(os.path.dirname(__file__) + "./../files/example.fastq")
```
tjanez: Is there a reason to prefer `os.path.normpath` over more frequently used `os.path.abspath`?

Also the `./` part in the file name is redundant?

resdk/tests/unit/test_resolwe.py
```diff
@@ -443,7 +450,8 @@ def setUp(self, auth_mock):  # pylint: disable=arguments-differ
     def test_bad_url(self, requests_mock):
         requests_mock.post = MagicMock(side_effect=[requests.exceptions.ConnectionError()])
 
-        with six.assertRaisesRegex(self, ValueError, r"Server not accessible on .*. Wrong url?"):
+        with six.assertRaisesRegex(self, ValueError,
+                                   'Server not accessible on www.abc.com. Wrong url?'):
```
tjanez: This is inconsistent, the string is not a regular expression anymore, but the error is still checked with `six.assertRaisesRegex`. What do we want?

setup.py
```diff
@@ -62,10 +62,11 @@
         'test': [
             'check-manifest',
             'mock==1.3.0',
-            'pep8>=1.6.2',
+            'pycodestyle>=2.0.0',
+            'pydocstyle>=1.0.0',
             'pylint>=1.4.3',
```
tjanez: Maybe also bump the pylint version since we probably aren't sure whether everything works on older versions?

docs/CHANGELOG.rst
```diff
@@ -5,6 +5,15 @@ Change Log
 All notable changes to this project are documented in this file.
 
 
+==========
+Unreleased
+==========
+
+Added
+-----
+* Check PEP 8
```
tjanez: Please, add a note that PEP 257 is also being checked.

resdk/resolwe.py
```diff
@@ -70,7 +71,7 @@ def _version_string_to_int(self, version):
         version_numbers = [int(number_string) for number_string in version.split(".")]
 
         if len(version_numbers) > len(VERSION_NUMBER_BITS):
-            raise NotImplementedError("Versions with more than {0} decimal places are not supported.".format(
+            raise NotImplementedError("Version should have at most {0} decimal places.".format(
```
tjanez: The `0` from `{0}` can be removed.

resdk/resources/base.py
```diff
@@ -105,4 +102,6 @@ def update(self):
         self._update_fields(self.api(self.id).get())
 
     def __repr__(self):
-        return u"{} <id: {}, slug: '{}', name: '{}'>".format(self.__class__.__name__, self.id, self.slug, self.name)
+        """Format resource name."""
+        return u"{} <id: {}, slug: '{}', name: '{}'>".format(self.__class__.__name__,
```
tjanez: Drop the `u` from `u"{}..."` since all strings will automatically be unicode due to `from __future__ import unicode_literals`.

resdk/resolwe.py
```diff
@@ -70,7 +71,7 @@ def _version_string_to_int(self, version):
         version_numbers = [int(number_string) for number_string in version.split(".")]
```
tjanez: On a more general note, this function looks like it was copied from [django-versionfield package](https://github.com/tonioo/django-versionfield/blob/master/versionfield/utils.py#L7-L40). Why not just require that package?

resdk/tests/unit/test_resolwe.py
mstajdohar: You should ask the author. Here I have only removed warnings from the original implementation. I am now refactoring the original implementation to use a constant BASE_DIR where I use `abspath`:
```
BASE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
```

resdk/resolwe.py
mstajdohar: I did not want a dependency to anything related to Django in the SDK. The django-versionfield does not define a Django requirement in the `setup.py`, but it should. It imports modules from Django. I suggest to make the SDK as lean as possible.

resdk/tests/unit/test_resolwe.py
mstajdohar: This is how you check literals as well. Look at the example of [Python's assertRaisesRegexp here](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp). We use the implementation in six to be compatible with py2 and py3.

mstajdohar: We wish to check this exact string (not the regex), because this exact url is used in this test case.

mstajdohar:
All issues have been addressed. Merging to master.

PR #41, opened 2016-09-16
Add Feature resource and import script

kostko:
Requires genialis/resolwe-bio#107.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -13,6 +13,7 @@ Added
 -----
 * Check PEP 8 and PEP 257
 * Support delete() method
+* Feature resource and resolwe-update-kb script
 
 Changed
 -------
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -22,6 +22,7 @@ Changed
 Fixed
 -----
 * Fix resolwe-upload-reads script
+* Rename ResolweQuerry to ResolweQuery
 
 
 ==================
@@ -101,7 +102,7 @@ Changed
 Added
 -----
 * Endpoints ``data``, ``sample`` and ``collections`` in ``Resolwe`` class
-* ``ResolweQuerry`` class with ``get`` and ``filter`` methods
+* ``ResolweQuery`` class with ``get`` and ``filter`` methods
 * ``Sample`` class with ``files`` and ``download`` methods
 * Tox configuration for running tests
 * Travis configuration for automated testing
@@ -110,7 +111,7 @@ Changed
 -------
 * Rename resolwe_api to resdk
 * Add ``data``, ``sample``, ``collections`` to ``Resolwe`` class and create
-  ``ResolweQuerry`` class
+  ``ResolweQuery`` class
 * Move ``data.py``, ``collections.py`` ... to ``resources`` folder
 * Remove ``collection``, ``collection_data`` and ``data`` methods from
   ``Resolwe`` and from tests.
--- a/resdk/resolwe.py
+++ b/resdk/resolwe.py
@@ -7,7 +7,7 @@ Resolwe
 .. autoclass:: resdk.Resolwe
    :members:
 
-.. autoclass:: resdk.ResolweQuerry
+.. autoclass:: resdk.ResolweQuery
    :members:
 
 """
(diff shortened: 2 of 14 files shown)


resdk/scripts.py
```diff
@@ -350,3 +352,85 @@ def upload_reads():
...
+    parser.add_argument('features', help='ZIP file containing feature descriptions')
+
+    args = parser.parse_args()
+
+    if args.verbose:
+        resdk_logger.start_logging()
+
+    resolwe = Resolwe(args.email, args.password, args.address)
+
+    subtype_map = {
```
kostko: @jkokosar this should be a mapping between what is found in the files being imported and the [internal designations](https://github.com/genialis/resolwe-bio/pull/107/files#diff-d65dd8283ab5cdf96b0e933ba90a5bcfR15). Everything, which is not in this list will be imported as `other`, so can you check how these types should map and update this dictionary?

resdk/scripts.py
```diff
@@ -350,3 +352,85 @@ def upload_reads():
...
+        'processed_pseudogene': 'pseudo',
+        'antisense': 'asRNA',
+        'protein_coding': 'protein-coding',
+        # SUBTYPE_PROTEIN_CODING = 'protein-coding'
+        # SUBTYPE_PSEUDO = 'pseudo'
+        # SUBTYPE_RRNA = 'rRNA'
```
kostko: @jkokosar this should be a mapping between what is found in the files being imported and the [internal designations](https://github.com/genialis/resolwe-bio/pull/107/files#diff-d65dd8283ab5cdf96b0e933ba90a5bcfR15). Everything, which is not in this list will be imported as `other`, so can you check how these types should map and update this dictionary?

kostko:
Looks like linters tox is failing, because it doesn't take the updated `.pylintrc` into account (I've updated it as it was inconsistent with other Resolwe repositories, `max-line-length` was 99 instead of `119`).

resdk/resolwe.py
```diff
@@ -494,7 +497,7 @@ def __init__(self, resolwe, Resource, endpoint=None):
         self.resolwe = resolwe
         self.resource = Resource
         self.endpoint = endpoint if endpoint else Resource.endpoint
-        self.api = getattr(resolwe.api, self.endpoint)
+        self.api = operator.attrgetter(self.endpoint)(resolwe.api)
```
JureZmrzlikar: What is the benefit of using `operator`?

resdk/scripts.py
```diff
@@ -350,3 +352,122 @@ def upload_reads():
...
+
+    errors = 0
+    updated = 0
+    with zipfile.ZipFile(args.features) as archive:
+        for entry in archive.infolist():
+            if not entry.filename.endswith('.tab'):
+                continue
+
+            if args.verbose:
+                print('Importing features from "{}"...'.format(entry.filename))
```
JureZmrzlikar: Since `verbose` starts logging, it would probably make sense to replace `print` statements with logger calls?

resdk/resolwe.py
kostko: It can resolve nested paths like `kb.feature.admin`, while `getattr` would require three nested calls to achieve the same (eg. `getattr(getattr(getattr(x, 'kb'), 'feature'), 'admin')`).

JureZmrzlikar: Ah, OK. Nice solution.

mstajdohar:
Please improve the test coverage.

docs/CHANGELOG.rst
```diff
@@ -101,7 +103,7 @@ Changed
 Added
 -----
 * Endpoints ``data``, ``sample`` and ``collections`` in ``Resolwe`` class
-* ``ResolweQuerry`` class with ``get`` and ``filter`` methods
+* ``ResolweQuery`` class with ``get`` and ``filter`` methods
```
mstajdohar: +1

resdk/scripts.py
```diff
@@ -350,3 +352,122 @@ def upload_reads():
...
+                    print("WARNING: Failed to update feature '{}:{}':\n{}\n{}".format(
+                        row['Source'],
+                        row['ID'],
+                        row,
+                        error.response.content  # pylint: disable=no-member
+                    ))
+                    errors += 1
+
+    if errors:
+        print("Encountered {} errors during import.".format(errors))
```
mstajdohar: Should we use logger like in the rest of the SDK?

kostko: Yes, I will fix this.

kostko: Checked this again, the current scripts don't use logger, but `print` directly? See `sequp` for example.

JureZmrzlikar: True. But we plan(ed) to implement it... If you want, you can probably leave it as it is, and this can be solved in separate PR.

kostko: No, I will update it everywhere to use logger :-)

mstajdohar: > Checked this again, the current scripts don't use logger, but print directly? See sequp for example.

They should be refactored. We introduced logger after the first scripts.

kostko: The current linter rules are really in the way:
```
C:104, 0: Invalid constant name "logger" (invalid-name)
W:152,16: Use % formatting in logging functions and pass the % parameters as arguments (logging-format-interpolation)
```

In the `resdk_loggers` documentation using the `logger` variable name is recommended and we are also using `format` for interpolation everywhere.

mstajdohar: We use [BraceMessage](https://github.com/genialis/resolwe/blob/master/resolwe/utils.py#L4) in resolwe and resolwe-bio to avoid the formatting issue. See a usage example [here](https://github.com/genialis/resolwe/blob/3e4bc3211ccc03eca79647ca988bd0afb91b3b25/resolwe/flow/managers/base.py#L22). You can add an exception for the invalid constant name. Can you tell pylint that logger is an OK constant?

PR #42, opened 2016-09-30
Save resource to server

mstajdohar:


mstajdohar:
Still need to improve the implementation and write unit tests. How do you like it conceptually?

resdk/resources/base.py
```diff
@@ -87,19 +90,57 @@ def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
 
         self.logger = logging.getLogger(__name__)
 
-    def _update_fields(self, fields):
+    def _add_fields(self, fields_list_type, field_names):
```
mstajdohar: Maybe we should call this `_register_fields`.

resdk/resources/base.py
```diff
@@ -28,37 +28,40 @@ class BaseResource(object):
...
                 raise ValueError("id should be an integer.")
         elif slug:
             if not isinstance(slug, six.string_types):
                 raise ValueError("slug should be a string.")
-        else:
+        elif model_data:
             if not isinstance(model_data, dict):
                 raise ValueError("model_data should be a dict.")
 
+        self._add_fields('_writable_fields', ('slug', 'name', 'permissions'))
```
mstajdohar: Maybe we should define this (_writable_fields, _update_protected_fields, _read_only_fields) as class attributes. The class instance should not change this lists of fields.

resdk/resources/base.py
```diff
@@ -26,39 +26,43 @@ class BaseResource(object):
 
     endpoint = None
 
+    WRITABLE_FIELDS = ('slug', 'name', 'permissions')
```
mstajdohar: I have made the lists of fields class constants. How do you like the new solution?

resdk/resources/base.py
```diff
@@ -87,19 +91,49 @@ def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
...
+        response = self.api(self.id).get()
+        self._update_fields(response)
+
+    def save(self):
+        """Save resource to the server."""
+        if self.id:  # update resource
+            payload = {}
+            for field_name in self.WRITABLE_FIELDS:
+                # find changes
+                if getattr(self, field_name) != self._original_values.get('field_name'):
```
JureZmrzlikar: `'field_name'` should probably be without Apostrophe?

JureZmrzlikar: `'field_name'` should probably be without Apostrophe?

resdk/resources/base.py
```diff
@@ -87,19 +91,49 @@ def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
...
+                    payload[field_name] = getattr(self, field_name)
+
+            if payload:
+                response = self.api(self.id).patch(payload)
```
JureZmrzlikar: Surround with `try-catch`. Request can fail for several reasons, for example when validating the updated descriptor. Probably we wish to print descriptive message what went wrong: server unresposive, bad request, server error etc...

JureZmrzlikar: Also, adding and removing data objects to/from collections can only be done via `add_data` and `remove_data` requests. Exmaple:
`res.api.collection(id).add_data.post({'ids': [1, 2, 3]})`

resdk/resources/base.py
```diff
@@ -87,19 +91,49 @@ def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
...
         """Update resource fields from the server."""
-        self._update_fields(self.api(self.id).get())
+        response = self.api(self.id).get()
+        self._update_fields(response)
+
+    def save(self):
+        """Save resource to the server."""
+        if self.id:  # update resource
+            payload = {}
+            for field_name in self.WRITABLE_FIELDS:
```
JureZmrzlikar: If user has changed some fields that are and some that are not writable, do we wish to inform him that only part of his changes were submitted?

resdk/resources/base.py
mstajdohar: Fixed

resdk/resources/base.py
mstajdohar: > Surround with try-catch. Request can fail for several reasons, for example when validating the updated descriptor. Probably we wish to print descriptive message what went wrong: server unresposive, bad request, server error etc...

This is a library and I want those exception to be raised directly, so that we can handle them on higher levels.

mstajdohar: > Also, adding and removing data objects to/from collections can only be done via add_data and remove_data requests. 

I think my implementation is sufficient for the next release. We should support this functions directly in the collection resource (BaseCollection?) soon, though. Please add this task to the spec.

mstajdohar:
Added tests, please do a final review pass.

resdk/resources/base.py
dblenkus: We had similar discussion about permissions on resolwe (what happens if user only has permissions for one part of the request) and final decision was that whole request gets rejected. Should we follow similar logic here?

resdk/resources/base.py
dblenkus: It is dangerous to override some (just created) data on the server if you use `PATCH` instead of `add_data`/`remove_data`. It's not a problem while we don't have much users, but please at least add comment to fix it.

resdk/resources/base.py
kostko: Yes, I think that if a field is not writable and the user attempts to change it, it should be rejected.

> I would have to somehow distinguish the changes from the server and users.

You can track dirty fields by using [descriptors](https://docs.python.org/2.7/howto/descriptor.html) for them, as long as they are scalars (eg. not lists/dicts), set using operator `=` or `setattr`. If you want to track modifications to non-scalar fields, you need some additional logic (a dirty-tracking list/dict/etc.).

mstajdohar: Done

mstajdohar:
Please have another review pass

PR #44, opened 2016-10-07
Fix sequp script descriptor to match the updated reads and sample descriptor schemas

jkokosar:


resdk/scripts.py
```diff
@@ -249,7 +249,7 @@ def parse_annotation_file(annotation_file):
 
                     if 'FASTQ_PATH' in row:
                         for seqfile in row['FASTQ_PATH'].split(','):
-                            seq_path = os.path.join(genialis_seq_dir, seqfile)
+                            seq_path = os.path.normpath(os.path.join(genialis_seq_dir, seqfile))
```
mstajdohar: Using `os.path.abspath` is generally safer. Read about the difference between abspath and normpath: http://www.diveintopython.net/functional_programming/finding_the_path.html

PR #45, opened 2016-10-14
Multiple mini fixes and improvements

dblenkus:


resdk/resolwe.py
```diff
@@ -293,16 +293,23 @@ def run(self, slug=None, input={}, descriptor=None,  # pylint: disable=redefined
...
 
-        # Upload files in basic:file fields
+        # Pre-process inputs
         try:
             for schema, fields in iterate_fields(input, process['input_schema']):
                 field_name = schema['name']
                 field_type = schema['type']
                 field_value = fields[field_name]
 
+                # Wrapp `list:` fields into list if they are not already
```
mstajdohar: Kind of nice, but maybe a bit hackish. Shouldn't users be aware that some field is a list field?

resdk/resources/data.py
```diff
@@ -146,6 +146,9 @@ def put_in_download_list(elm, fname):
             else:
                 raise KeyError("Item {} does not contain 'file' key.".format(fname))
 
+        if field_name and not field_name.startswith('output.'):
+            field_name = 'output.{}'.format(field_name)
```
mstajdohar: I like this one.

resdk/resolwe.py
dblenkus: I had the same thoughts, but came to the conclusion that user probably doesn't care what is happening in the background as long as things work and are nice to use.
There is really ugly exception if you try to run `res.run('upload-fastq-single', input={'src': 'reads.fastq.gz'})` only because you forgot to add the brackets.

mstajdohar: I agree that exception messages in resdk are terrible and should be improved. Our server is quite advanced and usually reports the error in the response body. 

I thought about this briefly and this is how I would solve the problem of ugly exceptions. We should create our exception class (in, e.g., `exceptions.py`), pass it the slumber/requests exception, extract the information from request body, append it to the exception message and raise our exception. [Here](https://github.com/genialis/resolwe-bio-py/blob/9f1aec60e5853cbe0cc1b2b8cb05b7ddeeb6eb94/resdk/resources/base.py#L154) is one example of how I made the exception message more informative. We could abstract this with a custom exception class.

Now to the second point, should we wrap values in lists automatically or not? I would like to hear more opinions, @jkokosar?

dblenkus: I agree that we should handle exceptions better. But first thing is to prevent exceptions from happening. Isn't our platform about taking care of tasks that we can instead of user?

mstajdohar: Now that I have slept on this, I think we should raise an error. We are building an SDK and exceptions are fine in libraries. I agree that for the user working in Python interpreter this is not convenient, but imagine a software package like Orange using our Python SDK. Exceptions should get raised for wrong inputs. Otherwise our SDK will get as messy as JavaScript :) Where you never know what evaluates to True and what not. What we want here is a ValidationError that is raised for wrong inputs. The validation can happen locally or on the server. If locally, it would work faster, but we will have to keep it in sync with the validation on the server. If on the server, we would have to capture the response, parse and reformat the exception.

dblenkus: > Exceptions should get raised for wrong inputs.

The inputs are not wrong, they are just not formatted in the right way :).

I agree that exception is better from Orange point of view. But on the other side, we should make it friendly for the users that are not top developers (or are not developers at all). So we must consider both sides.

In this case I don't see a problem wrapping value into list as it is clearly specified that list is required and all other options will fail. So there is no "javascript" magic in my opinion.

PR #47, opened 2016-10-18
Auto-hydrate data list in collection resource

dblenkus:
Method used for hydration is not optimal at the moment (it would be better to reload collection with `?hydrate_data=1` and manually generate list of Data objects), but enables some nice features in the future when queries will be lazy and compossible, i.e.: `collection.data.filter(type='data:differential-expression')` will load only DE objects

TODO:
- [x] write tests

### DIFF
--- a/resdk/resources/collection.py
+++ b/resdk/resources/collection.py
@@ -28,11 +28,14 @@ class BaseCollection(BaseResource):
     WRITABLE_FIELDS = ('data', 'description', 'settings', 'descriptor_schema',
                        'descriptor') + BaseResource.WRITABLE_FIELDS
 
+    #: ids of data objects (if not hydrated), data objects (if hydrated)
+    _data = None
+    #: indicates if ``_data`` list is already hydrated
+    _data_hydrated = False
+
     def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
                  model_data=None, resolwe=None):
         """Initialize attributes."""
-        #: id's of data objects in the resource
-        self.data = None
         #: a description
         self.description = None
         #: settings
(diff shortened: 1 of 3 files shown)


resdk/resources/collection.py
```diff
@@ -28,11 +28,14 @@ class BaseCollection(BaseResource):
     WRITABLE_FIELDS = ('data', 'description', 'settings', 'descriptor_schema',
                        'descriptor') + BaseResource.WRITABLE_FIELDS
 
+    #: id's of data objects in the resource
```
tjanez: Shouldn't this be "ids of data objects in the resource (if not hydrated), data objects in the resource (if hydrated)"

NOTE: Use "ids" instead of "id's"

dblenkus: Fixed

resdk/resources/collection.py
```diff
@@ -28,11 +28,14 @@ class BaseCollection(BaseResource):
     WRITABLE_FIELDS = ('data', 'description', 'settings', 'descriptor_schema',
```
mstajdohar: If I remember correctly you should remove the `'data'` from the `WRITABLE_FIELDS` list. Otherwise `BaseResource` expects this is a resource filed and may override the property.

mstajdohar: It may be confusing that getting the data would return hydrated objects whereas setting them expects ids. I suggest not to allow saving (to add or remove) data through the `data` attribute. Maybe we should even make this a method, i.e. `data()`. We should then support adding and removing data through helper functions, e.g. `add_data` and `remove_data`.

mstajdohar: I guess your tests did not cover all the bordering cases. What would happen now if you get the data, change one of the hydrated data objects and then call `save()` on the collection? Or if a user appends an ID to the collections of hydrated data objects and then calls `save()` on the collection? ...

mstajdohar: OK. Did you also remove it from the list of fields? It should not be treated as a field, because fields are handled by `.save()`, `.update()`... methods and this lazy field has other logic.

mstajdohar: Also, `.update()` should probably clear the cached values of lazy fields.

PR #48, opened 2016-10-19
Multiple fixes

dblenkus:


resdk/resources/data.py
```diff
@@ -145,7 +147,12 @@ def presample(self):
         # ``presample`` can not be cached as it can be converted to
         # ``sample``.
         presample = self.resolwe.presample.filter(data=self.id)
-        return None if len(presample) == 0 else presample[0]
+        if len(presample) == 0:
+            self._presample = None
+        # don't override existing sample object
```
tjanez: I would change this to "don't overwrite existing ..."

resdk/tests/unit/test_data.py
```diff
@@ -60,6 +60,14 @@ def test_presample(self):
         data.resolwe.presample.filter = MagicMock(return_value=['presample'])
         self.assertEqual(data.presample, 'presample')
 
+        # don't override existing presample
```
tjanez: Same as above.

resdk/resources/collection.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
     def data(self):
         """Lazy load ``data`` objects belonging to the collection."""
         if not self._data_hydrated:
-            self._data = self.resolwe.data.filter(id__in=','.join(map(str, self._data)))
+            self._data = self.resolwe.data.filter(id__in=','.join(map(str, self._data or [])))
```
tjanez: Wouldn't `[str(_) for _ in self._data]` instead of `map(str, self._data or []))` be more readable?

resdk/resources/utils.py
```diff
@@ -94,3 +96,35 @@ def endswith_colon(schema, field):
...
+
+def get_resource_id(resource):
+    """Return id attribute of the object if it is resorce, othervise return given value."""
+    return resource.id if isinstance(resource, BaseResource) else resource
+
+
+class resource_list(list):  # pylint: disable=invalid-name
+    """Subclass of ``list`` used for resdk resources.
+
+    This list is aware of resource objects in it and know how to compare
```
tjanez: I would change this "This list is aware of resource objects it holds and knows how to compare them with a list of just object ids."

resdk/resources/utils.py
```diff
@@ -94,3 +96,35 @@ def endswith_colon(schema, field):
     """
     if field in schema and not schema[field].endswith(':'):
         schema[field] += ':'
+
+
+def get_resource_id(resource):
```
tjanez: Is this a typical name for such functions? If not, another idea would be to use `get_resource_or_id` to be more explicit about what it does...

resdk/resources/collection.py
```diff
@@ -64,6 +68,16 @@ def data(self, value):
         self._data = value
         self._data_hydrated = False
 
+    def add_data(self, *data):
+        """Add ``data`` objects to the collection."""
+        data = [get_resource_id(d) for d in data]
+        self.api(self.id).add_data.post({'ids': data})
+
+    def remove_data(self, *data):
+        """Remove ``data`` objects to the collection."""
```
tjanez: "to the collection" -> "from the collection"

dblenkus:
Fixed

resdk/resources/utils.py
tjanez: Ok, then leave it as is.

resdk/resources/collection.py
dblenkus: You have to add `or []` to your example, as `self._data` is `None` when new object is created. So I prefer my way

tjanez: Ok, agreed.

resdk/resources/collection.py
```diff
@@ -64,6 +68,16 @@ def data(self, value):
         self._data = value
         self._data_hydrated = False
 
+    def add_data(self, *data):
```
mstajdohar: Nice. I like that you have supported this. Write in docs what can be the input parameter. We could/should? also support adding or removing data by slugs. This would probably be the most common use case when working in the interactive Python shell.

resdk/resources/utils.py
mstajdohar: This implementation makes it hard to support adding/removing data by slugs. The `get_resource_id` function does not know what resource to query.

resdk/resources/collection.py
mstajdohar: I don't know that is a bit hackish too. Let's sleep on this and plan it next year.

PR #50, opened 2016-10-21
Fix files  and download method in BaseCollection

dblenkus:


resdk/resources/collection.py
```diff
@@ -124,20 +122,11 @@ def download(self, file_name=None, file_type=None, download_dir=None):
         if file_type and not isinstance(file_type, six.string_types):
             raise ValueError("Invalid argument value `file_type`.")
 
-        for id_ in self.data:
-            data = Data(id=id_, resolwe=self.resolwe)
+        for data in self.data:
+            data_files = data.files(file_name, file_type)
+            files.extend(zip([data.id] * len(data_files), data_files))
```
mstajdohar: The zip here and formatting later logic is quite unreadable. Why not change it to:
```
files.extend('{}/{}'.format(data.id, file_name) for file_name in data_files))
```

And remove line 129.

PR #52, opened 2016-10-21
Add get_or_run method

dblenkus:


resdk/resolwe.py
```diff
@@ -243,7 +243,7 @@ def _process_file_field(self, path):
 
     def run(self, slug=None, input={}, descriptor=None,  # pylint: disable=redefined-builtin
             descriptor_schema=None, collections=[],
-            data_name='', src=None, tools=None):
+            data_name='', src=None, tools=None, use_existing=False):
```
dblenkus: @mstajdohar: If `use_existing` is set to `True` and matching object exists, all other parameters (i.e. `collections`, `descriptor`,..) are ignored. Should I add a check that makes sure that other parameters are not given in this case?

mstajdohar: If given, check that they match the returned object. We should find a better parameter name for `use_existing`.

dblenkus: And what to do if they don't match? Create new object with given parameters or raise an error? I don't like neither of these two options :)

Do you have any suggestions for better name?

dblenkus:
@mstajdohar: Updated as discusses. Please rereview.

resdk/resolwe.py
```diff
@@ -241,6 +242,63 @@ def _process_file_field(self, path):
             'file_temp': file_temp,
         }
 
+    def _get_process(self, slug=None):
+        """Return process with given slug.
+
+        Raise error if process doesn't exist or more than one is returned
```
mstajdohar: Add a full stop to the end of the sentence.

resdk/resolwe.py
```diff
@@ -241,6 +242,63 @@ def _process_file_field(self, path):
...
+            process = process[0]
+        elif len(process) == 0:
+            raise ValueError("Could not get process for given slug.")
+        else:
+            raise ValueError("Unexpected behaviour at get process with slug {}".format(slug))
+
+        return process
+
+    def _process_inputs(self, inputs, process):
+        """Process input fields."""
```
mstajdohar: Write a paragraph about how do you change the inputs.

resdk/resolwe.py
```diff
@@ -241,6 +242,63 @@ def _process_file_field(self, path):
...
+                    fields[field_name] = self._process_file_field(field_value)
+
+                # Upload files in list:basic:file` fields
+                elif field_type == 'list:basic:file:':
+                    fields[field_name] = []
+                    for obj in field_value:
+                        fields[field_name].append(self._process_file_field(obj))
+
+        except KeyError as key_error:
+            raise KeyError("Field '{}' not in process '{}' input schema.".format(key_error.args[0],
```
mstajdohar: Should this be a ValidationError?

resdk/resolwe.py
```diff
@@ -241,6 +242,63 @@ def _process_file_field(self, path):
...
+        inputs = copy.deepcopy(inputs)  # leave original intact
+
+        try:
+            for schema, fields in iterate_fields(inputs, process['input_schema']):
+                field_name = schema['name']
+                field_type = schema['type']
+                field_value = fields[field_name]
+
+                # Wrap `list:` fields into list if they are not already
+                if field_type.startswith('list:') and not isinstance(field_value, list):
```
mstajdohar: We should support this on the server. If you insist on this here, add an XXX comment to remove this when we support it on the server.

resdk/resolwe.py
```diff
@@ -241,6 +242,63 @@ def _process_file_field(self, path):
...
+                    fields[field_name] = [field_value]
+                    field_value = fields[field_name]  # update value for the rest of the loop
+
+                # Dehydrate `data:*` fields
+                if field_type.startswith('data:'):
+                    fields[field_name] = get_data_id(field_value)
+
+                # Dehydrate `list:data:*` fields
+                elif field_type.startswith('list:data:'):
+                    fields[field_name] = []
```
mstajdohar: This would be shorter and more readable:
```
fields[field_name] = [get_data_id(data) for data in field_value]
```

resdk/resolwe.py
```diff
@@ -241,6 +242,63 @@ def _process_file_field(self, path):
...
+                    for data in field_value:
+                        fields[field_name].append(get_data_id(data))
+
+                # Upload files in `basic:file` fields
+                elif field_type == 'basic:file:':
+                    fields[field_name] = self._process_file_field(field_value)
+
+                # Upload files in list:basic:file` fields
+                elif field_type == 'list:basic:file:':
+                    fields[field_name] = []
```
mstajdohar: Same here:
```
fields[field_name] = [self._process_file_field(obj) for obj in field_value]
```

resdk/resolwe.py
```diff
@@ -284,50 +342,8 @@ def run(self, slug=None, input={}, descriptor=None,  # pylint: disable=redefined
...
-                # Upload files in list:basic:file` fields
-                elif field_type == 'list:basic:file:':
-                    fields[field_name] = []
-                    for obj in field_value:
-                        fields[field_name].append(self._process_file_field(obj))
-
-        except KeyError as key_error:
-            raise KeyError("Field '{}' not in process '{}' input schema.".format(key_error.args[0],
-                                                                                 process['slug']))
+        process = self._get_process(slug)
```
mstajdohar: I like how you've refactored this (and the next line).

resdk/resolwe.py
dblenkus: `ValidationError` is Django specific, it is not defined in Python
https://docs.python.org/3/library/exceptions.html

dblenkus:
Fixed.

mstajdohar:
I would define our ValidationError in the `resdk/exceptions.py` module. Similar to how [Slumber defines their exceptions](https://github.com/samgiles/slumber/blob/master/slumber/exceptions.py).

dblenkus:
Added

PR #53, opened 2016-10-22
Add relations between resources

dblenkus:
Includes #51

### DIFF
--- a/resdk/resources/collection.py
+++ b/resdk/resources/collection.py
@@ -4,7 +4,7 @@ from __future__ import absolute_import, division, print_function, unicode_litera
 import six
 
 from .base import BaseResource
-from .utils import get_resource_id, resource_list
+from .utils import get_data_id, get_sample_id
 
 
 class BaseCollection(BaseResource):
@@ -26,12 +26,6 @@ class BaseCollection(BaseResource):
 
     WRITABLE_FIELDS = ('description', 'settings', 'descriptor_schema',
                        'descriptor') + BaseResource.WRITABLE_FIELDS
-    READ_ONLY_FIELDS = ('data', ) + BaseResource.READ_ONLY_FIELDS
-
-    #: ids of data objects (if not hydrated), data objects (if hydrated)
-    _data = None
-    #: indicates if ``_data`` list is already hydrated
-    _data_hydrated = False
 
     def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
                  model_data=None, resolwe=None):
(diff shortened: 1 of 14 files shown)


resdk/resources/sample.py
```diff
@@ -34,6 +41,25 @@ def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
...
 
+    @property
+    def collections(self):
+        """Lazy load `collection` objects to which sample belongs."""
+        if not self._collections_hydrated:
+            collections = self.resolwe.collection.filter(
+                id__in=','.join(map(str, self._collections or [])))
+            # Transform list to ``resource_list``, so the comparison with the
+            # original value will work
+            self._collections = resource_list(collections)
```
dblenkus: Remove this when #54 is accepted.

dblenkus:
@mstajdohar Ready for review.

resdk/resources/collection.py
```diff
@@ -44,37 +41,31 @@ def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
...
         #: descriptor schema
         self.descriptor_schema = None
-        #: collections
-        self.collections = None
 
-        BaseResource.__init__(self, slug, id, model_data, resolwe)
+        super(BaseCollection, self).__init__(slug, id, model_data, resolwe)
+
+    def update(self):
+        """Clear cache before updating."""
```
mstajdohar: First write about what this function does in general (copy from BaseResource). Write specifics below.

resdk/resources/collection.py
```diff
@@ -152,11 +143,45 @@ class Collection(BaseCollection):
...
+        self._samples = None
+
+        super(Collection, self).update()
+
+    def add_samples(self, *samples):
+        """Add `samples` objects to the collection."""
+        # Prevent circular import
+        from .sample import get_sample_id
```
mstajdohar: Ugly. Why not have this in utilities?

resdk/resources/collection.py
```diff
@@ -152,11 +143,45 @@ class Collection(BaseCollection):
...
+
+        samples = [get_sample_id(s) for s in samples]
+        for sample in samples:
+            self.resolwe.api.sample(sample).add_to_collection.post({'ids': [self.id]})
```
mstajdohar: We should support adding a list of samples to a collection on the server. Maybe by adding an endpoint to the collection resource and call it `self.api. add_samples({'ids': [sample ids]})`.

resdk/resources/collection.py
```diff
@@ -152,11 +143,45 @@ class Collection(BaseCollection):
...
+
+    def remove_samples(self, *samples):
+        """Remove ``sample`` objects from the collection."""
+        # Prevent circular import
+        from .sample import get_sample_id
+
+        samples = [get_sample_id(s) for s in samples]
+        for sample in samples:
+            self.resolwe.api.sample(sample).remove_from_collection.post({'ids': [self.id]})
```
mstajdohar: Consider the above comment.

resdk/resources/sample.py
```diff
@@ -32,7 +35,21 @@ def __init__(self, slug=None, id=None,  # pylint: disable=redefined-builtin
 
         self.presample = presample
 
-        BaseCollection.__init__(self, slug, id, model_data, resolwe)
+        super(Sample, self).__init__(slug, id, model_data, resolwe)
+
+    def update(self):
+        """Clear cache before updating."""
```
mstajdohar: The first line should describe the functionality. Specifics should follow below.

resdk/resources/collection.py
dblenkus: `get_sample_id` needs to import `Sample` to check if given object is sample. If i put it in utilities there will be a lot of circular imports.
I think that we cannot solve this without circular imports (`Sample` inherits from `Collection` and we want to add `Sample` to the `Collection`).

resdk/resources/collection.py
dblenkus: I agree, main problem is that `Collection` is implemented in `resolwe` and `Sample` in `resolwe-bio`. We have to think how to handle such cases in general.

For now I Can add a comment that this has to be improved.

resdk/resources/collection.py
mstajdohar: What about if we check for
```
type(obj).__name__ == 'Sample'
```
or simply
```
if hasattr('id', obj):
    return getattr('id', obj)
...
```

resdk/resources/collection.py
mstajdohar: Good enough.

resdk/resources/collection.py
dblenkus: I prefer the first idea. The second one is dangerous, i.e. if you pass `Collection` instead of `Data` it will pass and modify random Data object.

dblenkus:
Fixed.

PR #54, opened 2016-10-22
Make ResolweQuery lazy

dblenkus:
**Includes:** #53 

Resolwe query is now lazy and composable, which means that multiple filters can be applied before real request is made.

At the moment, the main advantage of this approach is that i.e. `collection.data` returns query with already applied filter and it can be filtered further:

```
collection.data.filter(type='data:alignment:bam')
```

Another useful feature is that list slicing is supported and transformed to limit&offset pagination on the API, so `collection.data[5:10]`query will transform to the request `...&offset=5&limit=5`

**TODO:**
- [x] `collection` cannot be saved, because `data` field is considered as changed
- [x] `.get()` makes 2 identical requests
- [x] docstrings
- [x] tests


resdk/resources/base.py
```diff
@@ -118,7 +120,10 @@ def save(self):
         """Save resource to the server."""
         def field_changed(field_name):
             """Check if local field value is different from the server."""
-            return getattr(self, field_name) != self._original_values.get(field_name)
+            field_value = getattr(self, field_name)
+            if isinstance(field_value, ResolweQuery):
+                return False
```
dblenkus: @mstajdohar: What do you think about this solution? We cannot compare values if we don't execute query (what we obviously don't want to do)

resdk/query.py
```diff
@@ -0,0 +1,246 @@
...
+
+    A Resolwe instance (for example "res") has several endpoints:
+    res.data, res.collections, res.sample and res.process. Each
+    andpooint is an instance of the ResolweQuery class. ResolweQuery
+    supports querries on corresponding objects, for example:
+
+    res.data.get(42)  # return Data object with ID 42.
+    res.sample.filter(contributor=1)  # return all samples made by contributor 1
+
+    This
```
mstajdohar: Typo?

resdk/query.py
```diff
@@ -0,0 +1,246 @@
...
+
+    Return a list of Data objects that match kwargs.
```
mstajdohar: ... that match the given arguments.

resdk/query.py
```diff
@@ -0,0 +1,246 @@
...
+        new_query._offset = self._offset + index if self._offset else index
+        new_query._limit = 1
+
+        query_list = list(new_query)
+        if len(query_list) == 0:
+            raise IndexError('list index out of range')
+        return query_list[0]
+
+    def __iter__(self):
+        """Return itterator over current the object."""
```
mstajdohar: typo: over the current object

resdk/query.py
```diff
@@ -0,0 +1,246 @@
...
+        self._fetch()
+        return iter(self._cache)
+
+    def __repr__(self):
+        """Return string representation of the current object."""
+        self._fetch()
+        return '[{}]'.format(',\n '.join(str(obj) for obj in self._cache))
+
+    def __len__(self):
+        """Return length of results og current query."""
```
mstajdohar: typo: of

resdk/query.py
```diff
@@ -0,0 +1,246 @@
...
+        self._fetch()
+        return len(self._cache)
```
mstajdohar: In not cached, could we just query for the number of objects on the server instead of transferring down all objects?

resdk/query.py
```diff
@@ -0,0 +1,246 @@
...
+        return self.api.post(data)  # pylint: disable=no-member
+
+    def filter(self, **filters):
+        """Return clone of current query with added given filters."""
+        new_query = self._clone()
+        new_query._add_filter(filters)  # pylint: disable=protected-access
+        return new_query
+
+    def all(self):
+        """Return current queryset.
```
mstajdohar: typo: the current

resdk/query.py
```diff
@@ -0,0 +1,246 @@
...
+
+        This is handy function to get newly created query without any
```
mstajdohar: typo: This is a handy...

resdk/resolwe.py
```diff
@@ -156,8 +153,7 @@ def _register(self, src, slug):
             process['persistence'] = persistence_map[process['persistence']]
 
         try:
-            server_process = self.process.filter(slug=process['slug'],
-                                                 ordering='-version', limit=1)
+            server_process = self.process.filter(slug=process['slug'], ordering='-version')[:1]
```
mstajdohar: Are you sure this only fetches one object from the server? Excellent if it does.

resdk/query.py
dblenkus: This is the docstring that hasn't been finished yet :)

resdk/resolwe.py
dblenkus: Yes, I've tested it and it works.

dblenkus: It's just a bit ugly to have `ordering` in `filter`. Maybe we can add `.order_by` method in the future. And also `.first()` and `.last()` would be useful.

resdk/query.py
mstajdohar: Finish it or write a comment :)

resdk/query.py
dblenkus: API doesn't support this functionality. But it would be nice to have it :)

dblenkus:
@mstajdohar I've added another commit with final changes.

PR #56, opened 2016-11-01
Set chunk size for uploading files to 8MB

dblenkus:


mstajdohar:
Can you test how the chunk size affects the upload speed for very large files? Is 1MB optimal? I remember we experimented with this on genesis because Garvan had uploads over 120GB. On the genesis server we have `client_max_body_size    50M;`.

dblenkus:
`client_max_body_size` is setting in Nginx that determine max body size, it doesn't influence chunks in any way (except limiting their size).
I think that 1MB is universal setting that will cover most use-cases. But we should make this setting configurable (in resdk) and probably increase `client_max_body_size` in Nginx. @tjanez
I can do some testing, but it will take some time to setup environment and we must determine in what environment it should be done, i.e. local network, fast internet,...

mstajdohar:
I know. I am pretty sure the chunks were larger than 1 MB for the Garvan use case. I'll check.

mstajdohar:
We had chunk size 10 MB in the genesis-pyapi.

PR #60, opened 2016-11-03
Update authentication

dblenkus:


resdk/resolwe.py
```diff
@@ -37,9 +37,7 @@
 
 VERSION_NUMBER_BITS = (8, 10, 14)
 CHUNK_SIZE = 8000000  # 8MB
-DEFAULT_EMAIL = 'anonymous@genialis.com'
-DEFAULT_PASSWD = 'anonymous'
-DEFAULT_URL = 'https://dictyexpress.research.bcm.edu'
+DEFAULT_URL = 'http://localhost:8000'
```
dblenkus: @mstajdohar which server should be here? Maybe QA? But it is experimental and we don't guarantee that it works.

mstajdohar: QA is not OK. I guess eventually we will have a demo server, but for now we can have the localhost. Does the  `Resolwe()` function connect as an anonymous user if user name and password not given?

docs/CHANGELOG.rst
```diff
@@ -19,6 +19,8 @@ Changed
 * Include all necessary files for running the tests in source distribution
 * Exclude tests from built/installed version of the package
 * File field passed to ``run`` function can be url address
+* Default authentication settings are set to connect to local server as
```
mstajdohar: reformat: Connect to a local server as public user by default

resdk/resolwe.py
```diff
@@ -508,20 +512,26 @@ def _download_files(self, files, download_dir=None):  # pylint: disable=redefine
...
 
-    :param email: user's email
-    :type email: str
-    :param password: user's password
-    :type password: str
-    :param url: Resolwe server instance
-    :type url: str
+    :param str username: user's username
+    :param str password: user's password
+    :param str url: Resolwe server instance
```
mstajdohar: reformat: Resolwe server address

resdk/resolwe.py
```diff
@@ -508,20 +512,26 @@ def _download_files(self, files, download_dir=None):  # pylint: disable=redefine
...
+
+    def __init__(self, username=None, password=None, url=DEFAULT_URL):
         """Authenticate user on Resolwe server."""
         self.logger = logging.getLogger(__name__)
 
-        payload = {'username': email, 'password': password}
+        self.username = username
+        self.url = url
+
+        if not username and not password:
```
mstajdohar: Does this really work or should we get the anonymous user credentials?

dblenkus: It works, if you make request without cookies, public user is automatically assigned on the server.

dblenkus:
Fixed and added tests.

PR #61, opened 2016-11-06
Handle server errors in a better way

dblenkus:
This is the concept for handling errors in more verbose way. But some changes have to be 
made in Resolwe's responses to make this useful.

### DIFF
--- a/resdk/exceptions.py
+++ b/resdk/exceptions.py
@@ -11,6 +11,27 @@ Custom ReSDK exceptions.
 """
 from __future__ import absolute_import, division, print_function
 
+from slumber.exceptions import SlumberHttpBaseException
+
 
 class ValidationError(Exception):
     """An error while validating data."""
+
+
+class ResolweServerError(Exception):
+    """Error response from the Resolwe API."""
+
+
+def handle_http_exception(func):
+    """Decorator for handling slumber errors in more verbose way."""
+    def wrapper(*args, **kwargs):
+        """Transform slumber errors into ReSDK errors.
+
+        Use content of the HTTP response as exception error.
+        """
+        try:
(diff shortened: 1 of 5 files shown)


resdk/exceptions.py
```diff
@@ -11,6 +11,23 @@
...
 
 class ValidationError(Exception):
     """An error while validating data."""
+
+
+class ResloweAPIError(Exception):
+    """Error response from the Resolwe API."""
+
+
+def handle_slumber_exceptions(func):
```
mstajdohar: Consider renaming to handle_http_exception.

resdk/exceptions.py
```diff
@@ -11,6 +11,23 @@
...
 from __future__ import absolute_import, division, print_function
 
+from slumber.exceptions import SlumberHttpBaseException
+
 
 class ValidationError(Exception):
     """An error while validating data."""
+
+
+class ResloweAPIError(Exception):
```
mstajdohar: Consider renaming to ResloweServerError

resdk/resolwe.py
```diff
@@ -44,6 +44,18 @@
 TOOLS_REMOTE_HOST = os.environ.get('TOOLS_REMOTE_HOST', None)
 
 
+class ResolweResource(slumber.Resource):
```
mstajdohar: I worry this may be mistaken for resource classes in the resources module. It is probably OK, it fits nicely with the other ResolweXXX classes.

dblenkus: I can add `__all__ = ('Resolwe', )` at the beginning of the file to make it more obvious and prevent some errors.

dblenkus:
Updated

Issue #69, opened 2016-12-02
Helper functions for mapping gene IDs


mstajdohar:
Details in [spec](https://docs.google.com/document/d/1KOQjP1AoAXWb0MeVaL1Ss8-eAGyG9Ys-Wg6PrfBmuoU/edit#).

Write utility functions in resdk to facilitate the interaction with knowledge base backend. Each function should do no more than a single request to the knowledge base backend, improve the backend if necessary.

* A function that maps NCBI to MGI/Uniprot using a mapping table. Inputs should be: source_db, target_db and a list of feature_ids. Output is a list of mapped/converted feature_ids 
* A function that for a list of given features returns a dictionary of values from the gene info table. Inputs: source, a list of features and an optional parameter to limit which fields are returned (see gene info table/model for a list of valid options). Output is a dict {‘feature_id’: {‘source’: ‘NCBI, ‘organism’: ‘Homo sapiens’, … }}. Key ‘feature_id’ will always be unique since all the returned elements have the same source.

jkokosar:
@kostko When will you be able to continue work on this task? Other tasks (for the next milestone) are dependant on this.

kostko:
The backend part has already been done (and is merged), I only need to add support to the SDK. I will try to do this today.

kostko:
Done, see #72 for usage example. Note that this requires an update to be deployed to the backend first.

dblenkus:
@kostko can we close this?

PR #72, opened 2016-12-12
Add knowledge base feature/mapping lookup

kostko:
Requires genialis/resolwe-bio#162 deployed on the backend.

See #69.

Example use:
```python
# Get a list of mappings.
resolwe.mapping.filter(source_db='A', source_id__in=['foo', 'another'])
# Get a list of features.
resolwe.feature.filter(source='UCSC', query=['A1BG'])
```

### DIFF
--- a/resdk/query.py
+++ b/resdk/query.py
@@ -89,11 +89,19 @@ class ResolweQuery(object):
     api = None
     logger = None
 
-    def __init__(self, resolwe, Resource, endpoint=None):
+    def __init__(self, resolwe, resource, endpoint=None):
         """Initialize attributes."""
         self.resolwe = resolwe
-        self.resource = Resource
-        self.endpoint = endpoint if endpoint else Resource.endpoint
+        self.resource = resource
+
+        # Determine the endpoint to use.
+        if endpoint is not None:
+            self.endpoint = endpoint
+        elif resource.query_endpoint is not None:
+            self.endpoint = resource.query_endpoint
+        else:
+            self.endpoint = resource.endpoint
+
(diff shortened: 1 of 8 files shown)


jkokosar:
I have tested the functionality of the mapping/feature filters:

1) resolwe.mapping.filter()
 - is there a list of available filter parameters?
 - `source_id__in` can this be `query`?
 -  is there a way to limit which  'target_db' is returned?

2) resolwe.feature.filter()
 - is there a list of available filter parameters?
 - When I get the features retured using the filter(), how can I access the object fields (organism, description, aliases..)? Is this not yet supported?

jkokosar:
@dblenkus Please review

kostko:
> is there a list of available filter parameters?

Yes, at the moment you can query by `source_db`, `source_id`, `target_db` and `target_id`. Should we also allow filtering by `relation_type`?

> source_id__in can this be query?

Yes, we can change this.

> is there a way to limit which 'target_db' is returned?

Yes, querying by `target_db` should work.

> is there a list of available filter parameters?

Yes, currently `query` (genes) and `source`.

> When I get the features retured using the filter(), how can I access the object fields (organism, description, aliases..)? Is this not yet supported?

I've updated the branch, it should work now using attribute access.

kostko:
I'll update the backend so that mapping search is also handled by elastic search, so that we can have a separate search endpoint and a separate rest endpoint for updating mappings via the API.

jkokosar:
Should we also allow filtering by relation_type?

Not required at the moment.

kostko:
Updated the API. It now requires genialis/resolwe-bio#168 and genialis/genialis-base#192.

Issue #67, opened 2016-12-02
Permission denied error for logfile


mstajdohar:
If the SDK is installed using "sudo pip install resdk", the user gets msg "permission denied" for the logfile when loading the sdk using "import resdk".

This is not an issue if using python virtualenv.

mstajdohar:
Fixed

PR #73, opened 2016-12-15
Remove file logger

mstajdohar:
The file handler was not used. We should write the logger configuration from scratch in a better way when we will need to log to a file.


mstajdohar:
Closes #67.

PR #83, opened 2017-01-20
Support paginated responses in queries

dblenkus:


resdk/query.py
```diff
@@ -239,6 +239,10 @@ def _fetch(self):
             raise NotImplementedError(
                 'Unsupported query_method: {}'.format(self.resource.query_method))
 
+        # Extract data from paginated response
+        if len(items) == 4 and set(items) == set(['next', 'previous', 'count', 'results']):
```
kostko: This check looks strange. Why not check instead if `items` is a dictionary and it contains a `results` key?

dblenkus: To really make sure that response is paginated. If only check for `results` key is performed, strange errors can occurs if response contains `results` key by accident. But you are right, I have to add a check that `items` is a dict :)

kostko: How could it contain a `results` key by accident? I think this is something that is defined by the API and SDK should expect the server to return the correct response?

Also, if `items` contains a `results` key by accident, then `items` is still a dictionary and everything from there forward will be incorrect as it expects a list.

dblenkus: You are right :). I've fixed it

Issue #68, opened 2016-12-02
Download dir fields


mstajdohar:
One currently can not download fields of type DIR using the RESDK. Example is a "genome" object:

```
{
    u'fasta': {u'file': u'hg19.fasta.gz', u'size': 937047570},
    u'index_bt': {u'dir': u'bowtie_index', u'size': 3074126272},
    u'index_bt2': {u'dir': u'bowtie2_index', u'size': 7219579166},
    u'index_bwa': {u'dir': u'BWA_index', u'size': 5417472211},
    u'index_hisat2': {u'dir': u'hisat2_index', u'size': 4374656092},
    u'index_subread': {u'dir': u'subread_index', u'size': 5872181163}
}
```

dblenkus:
Closed in #91

PR #91, opened 2017-02-19
General improvements

### LINKED ISSUE #90: Queries in new objects are not evaluated correctly
When new object is created, all queries that are included (i.e. `collection.data`, `collection.samples`,...) are constructed in a wrong way, because object's id is not yet known and filters are applied with `id=None`.

Queries shouldn't be initialized in `__init__`, but on the first call and error should be raised if `id` is not yet known.

dblenkus:
Closes #90 and #68

### DIFF
--- a/resdk/query.py
+++ b/resdk/query.py
@@ -110,6 +110,7 @@ class ResolweQuery(object):
     """
 
     _cache = None
+    _count = None  # number of objects in current query (without applied limit and offset)
     _limit = None
     _offset = None
     _filters = collections.defaultdict(list)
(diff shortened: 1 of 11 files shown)


resdk/query.py
```diff
@@ -241,13 +237,31 @@ def _fetch(self):
 
         # Extract data from paginated response
         if isinstance(items, dict) and 'results' in items:
+            self._count = items['count']
             items = items['results']
 
-        self._cache = [self._populate_resource(data, **resource_inputs) for data in items]
+        self._cache = [self._populate_resource(data, resolwe=self.resolwe) for data in items]
```
mstajdohar: The `self._populate_resource` and `self.resolwe` are both members of this class. Why do you have to pass a class attribute to a class method? Couldn't you simply access `self.resolwe` in `_populate_resource`?

dblenkus: Good point :)

PR #93, opened 2017-02-28
Upload vaccinesurvey data

JureZmrzlikar:


mstajdohar:
Does this fit into resdk, @dblenkus, @jkokosar? Should we create a separate Python project?

dblenkus:
I imagine it would be more collection of the scripts than actual Python package: https://github.com/genialis/resdk-scripts

setup.py
```diff
@@ -72,6 +72,7 @@
             'resolwe-sequp = resdk.scripts:sequp',
             'resolwe-upload-reads = resdk.scripts:upload_reads',
             'resolwe-update-kb = resdk.scripts:update_knowledge_base',
+            'resolwe-upload-vcsrvy = resdk.scripts:upload_vaccinesurvey_data',
```
mstajdohar: Rename to `resolwe-upload-vaccinesurvey`

resdk/scripts.py
```diff
@@ -481,3 +484,74 @@ def update_knowledge_base():
...
+
+    strings = ['study_code', 'sex', 'ethnicity']
+    ints = ['village_code']
+    floats = ['body_temp']
+    dates = ['birth_date', 'entry_date']
+    booleans = ['fever', 'antimalaria_treatment', 'hospital_visit', 'vomit',
+                'cough', 'diarrhoea', 'bednet']
+    immunological = ['ama1', 'msp1', 'msp2', 'nanp', 'total_ige', 'haemoglobin']
+
+    with open(args.file, 'rt') as ofile:
```
mstajdohar: I suggest reading tab separated files with a [DictReader](https://docs.python.org/2/library/csv.html#csv.DictReader). It's more robust.

resdk/scripts.py
```diff
@@ -481,3 +484,74 @@ def update_knowledge_base():
...
+            try:
+                for col_name, col_value in zip(header, line):
+
+                    # Temporary fix for gender assignment (1=men, 2=women)
+                    if col_name == 'sex' and col_value in ['1', '2']:
+                        col_value = 'man' if col_value == '1' else 'woman'
+
+                    # Temporary fix if entry_date is missing
+                    if col_name == 'entry_date' and not col_value:
+                        col_value = datetime.now().date().strftime('%Y-%m-%d')
```
mstajdohar: Wouldn't be better to leave this empty if missing? You would have to update the schema so that this field would not be required.

resdk/scripts.py
```diff
@@ -481,3 +484,74 @@ def update_knowledge_base():
...
+                    if col_name in strings:
+                        descriptor['sample'][col_name] = col_value
+                    elif col_name in ints:
+                        descriptor['sample'][col_name] = int(col_value)
+                    elif col_name in floats:
+                        descriptor['sample'][col_name] = float(col_value)
+                    elif col_name in dates:
+                        descriptor['sample'][col_name] = col_value
+                    elif col_name in booleans:
+                        descriptor['sample'][col_name] = True if col_value == '1' else False
```
mstajdohar: You can simply write
```
descriptor['sample'][col_name] = col_value == '1'
```

JureZmrzlikar:
I will fix this when the `descriptor_schema` of real data will be known/decided. This PR has code that was written solely for demonstration purposes.

PR #95, opened 2017-03-17
Feature relations

dblenkus:


mstajdohar:
Looks good.

resdk/shortcuts/collection.py
```diff
@@ -0,0 +1,450 @@
...
+                relations[rel_type] = flatten_relations(relations[rel_type][None])
+            else:
+                relations[rel_type] = {
+                    key: flatten_relations(value)
+                    for key, value in six.iteritems(relations[rel_type])
+                }
+
+        relations = add_default_values(relations)
+
+        help_text = '''\
```
kostko: Perhaps this should be moved into a separate file, containing just the template, as this interrupts the code flow?

PR #97, opened 2017-03-27
Feature analysis

dblenkus:
Includes #95


resdk/analysis/_register.py
```diff
@@ -0,0 +1,28 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from resdk.resources import Collection, Relation, Sample
+
+from resdk.analysis.alignment import bowtie2, hisat2
+from resdk.analysis.chip_seq import macs, rose2
+from resdk.analysis.expressions import cuffnorm, cuffquant
+
+
+Collection.run_bowtie2 = bowtie2
```
mstajdohar: There must be a better way then `_register`. You do not have to find it now, but keep thinking about this.

resdk/resources/utils.py
```diff
@@ -114,3 +114,95 @@ def get_process_id(process):
...
+
+
+def get_relation_id(relation):
+    """Return id attribute of the object if it is relation, othervise return given value."""
+    return relation.id if type(relation).__name__ == 'Relation' else relation
+
+
+def is_collection(collection):
+    """Return ``True`` if passed object is Collection and ``False`` otherwise."""
+    return type(collection).__name__ == 'Collection'
```
mstajdohar: Why not use `isinstance`?

dblenkus: Because you cannot import `Collection` in to this file. Otherwise it causes cyclic import. It was your advice last time to do it this way :)

resdk/analysis/_register.py
```diff
@@ -0,0 +1,28 @@
+"""Patch ReSDK resources with analysis methods."""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from resdk.resources import Collection, Relation, Sample
+
+from resdk.analysis.alignment import bowtie2, hisat2
+from resdk.analysis.chip_seq import macs, rose2
+from resdk.analysis.expressions import cuffnorm, cuffquant
+
```
JenkoB: is bamplot missing? 
```from resdk.analysis.plots import bamplot```

File: resdk/analysis/_register.py [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+
+Collection.run_bowtie2 = bowtie2
+Collection.run_cuffnorm = cuffnorm
+Collection.run_cuffquant = cuffquant
+Collection.run_hisat2 = hisat2
+Collection.run_macs = macs
+Collection.run_rose2 = rose2
+
```
JenkoB: sae question here. How is with bamplot?

resdk/analysis/_register.py
dblenkus: @JenkoB thanks for spotting this! I've added it.

mstajdohar:
Fix linters.

PR #103, opened 2017-04-14
Fix error introduced by new pylint version

dblenkus:


mstajdohar:
Typo in commit message

tjanez:
Also, please pin the Pylint requirement to `~=1.7.0` in `setup.py` to avoid this and similar fallouts in the future.

PR #105, opened 2017-04-15
Order samples in relations in the same order as positions

dblenkus:


resdk/resources/relation.py
```diff
@@ -66,6 +66,8 @@ def samples(self):
         if not self._samples:
             sample_ids = [entity_obj['entity'] for entity_obj in self.entities]
             self._samples = self.resolwe.sample.filter(id__in=','.join(map(str, sample_ids)))
+            # Samples should be sorted, so they have same order as positions
+            self._samples = sorted(self._samples, key=lambda sample: sample_ids.index(sample.id))
```
mstajdohar: `sorted` will create a new array which is slower and takes extra space. What about calling `sort()` directly on `self._samples`?
```
self._samples.sort(key=lambda sample: sample_ids.index(sample.id))
```

PR #111, opened 2017-05-19
Remove labels input from cuffnorm

jvrakor:
@dblenkus @jkokosar @JenkoB 

Remove labels from cuffnorm in resdk. This is a follow-up to https://github.com/genialis/resolwe-bio/pull/271.

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -4,6 +4,14 @@ Change Log
 
 All notable changes to this project are documented in this file.
 
+==========
+Unreleased
+==========
+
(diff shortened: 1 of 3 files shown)


docs/CHANGELOG.rst
```diff
@@ -4,6 +4,14 @@ Change Log
 
 All notable changes to this project are documented in this file.
 
+==========
+Unreleased
+==========
+
+Fixed
```
dblenkus: Change this to `Changed` as this is not bug, but change of behaviour on backend.

PR #114, opened 2017-05-29
Minifixes

dblenkus:


mstajdohar:
Would be nice if we had tests for this function.

PR #115, opened 2017-05-30
Add cuffdiff helper function

jvrakor:
@dblenkus @jkokosar

Add cuffdiff helper function.

### DIFF
--- a/resdk/analysis/differential_expressions.py
+++ b/resdk/analysis/differential_expressions.py
@@ -1,0 +1,135 @@
+"""Differential expressions analysis."""
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from resdk.resources.utils import (
+    get_data_id, get_resolwe, get_resource_collection, get_samples, is_collection, is_relation,
+)
+
+__all__ = ('cuffdiff', )
+
+
+def cuffdiff(resource, annotation, genome=None, multi_read_correct=None, fdr=None,
+             library_type=None, library_normalization=None, dispersion_method=None, threads=None):
+    """Run Cuffdiff_ for selected cuffquants.
+
+    This method runs `Cuffdiff`_ process with ``annotation`` specified
+    in arguments. Library type is by defalt fr-unstranded. Other parameters
+    defaults: multi_read_correct=false, fdr=0.05, library_normalization=geometric,
+    dispersion_method=pooled, threads=1. Parameter genome is optional.
+
+    The way the function works depends on the resource. If it is run on a collection,
+    it will perform cuffdiff on every 'compare' relation labeled 'case-control' in
+    the selected collection. If it is run on a list of samples (not necesssarily in
+    the same collection) it will run cuffdiff on all 'compare' relations labeled
+    'case-control' containing all of the given samples but will discard those
+    samples in a relation that are not in the list of samples.
+
+    .. _Cuffdiff:
+        http://resolwe-bio.readthedocs.io/en/latest/catalog-definitions.html#process-cuffdiff
+
+    :param annotation: annotation file
+    :type annotation: `~resdk.resources.data.Data`
+    :param genome: genome object to use for bias detection and
+        correction algorithm
+    :type genome: `~resdk.resources.data.Data`
+    :param bool multi_read_correct: do initial estimation procedure to
+        more accurately weight reads with multiple genome mappings
+    :param fdr: the allowed false discovery rate
+    :type fdr: decimal
+    :param str library_type: options are: fr-unstranded, fr-firststrand,
+        fr-secondstrand
+    :param str library_normalization: options are: geometric, classic-fpkm,
+        quartile
+    :param str dispersion_method: options are: pooled, per-condition,
+        blind, poisson
+    :param int threads: use this many processor threads
+
+    """
+    inputs = {'annotation': get_data_id(annotation)}
+
+    input_objects = [annotation]
+
+    if genome is not None:
+        inputs['genome'] = genome
+        input_objects.append(genome)
+
+    if multi_read_correct is not None:
+        inputs['multi_read_correct'] = multi_read_correct
+
+    if fdr is not None:
+        inputs['fdr'] = fdr
+
+    if library_type is not None:
+        inputs['library_type'] = library_type
+
+    if library_normalization is not None:
+        inputs['library_normalization'] = library_normalization
+
+    if dispersion_method is not None:
+        inputs['dispersion_method'] = dispersion_method
+
+    if threads is not None:
+        inputs['threads'] = threads
+
+    samples = get_samples(resource)
+    sample_ids = [sample.id for sample in samples]
+
+    input_objects.extend(samples)
+    resolwe = get_resolwe(*input_objects)
+
+    collection_id = get_resource_collection(resource)
+
+    relation_filter = {}
+    if collection_id:
+        relation_filter['collection'] = collection_id
+    else:
+        relation_filter['entity'] = sample_ids
+
+    relations = resolwe.relation.filter(
+        type='compare',
+        label='case-control',
+        **relation_filter
+    )
+
+    cuffdiff_objects = []
+    for relation in relations:
+        control = []
+        case = []
+        for sample, position in zip(relation.samples, relation.positions):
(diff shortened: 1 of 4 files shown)


resdk/analysis/differential_expressions.py
```diff
@@ -0,0 +1,107 @@
...
+        if set(sample_ids_rel).issubset(sample_ids_res):
+            # maybe add a flag here and report error at the end corresponding to
+            # no relations containing selected samples.
+            for sample, position in zip(relation.samples, relation.positions):
+                if position == 'case':
+                    case.append(get_data_id(sample.get_cuffquant()))
+                elif position == 'control':
+                    control.append(get_data_id(sample.get_cuffquant()))
+                else:
+                    raise LookupError(
```
dblenkus: I think that `ValueError` would be more appropriate here.

https://docs.python.org/3/library/exceptions.html

resdk/analysis/differential_expressions.py
```diff
@@ -0,0 +1,107 @@
...
+        inputs['threads'] = threads
+
+    relation_filter = {}
+    collection_id = get_resource_collection(resource)
+    if collection_id:
+        relation_filter['collection'] = collection_id
+
+    samples = get_samples(resource)
+
+    input_objects = [annotation]
```
dblenkus: Also add `genome` to `input_objects`.

resdk/analysis/differential_expressions.py
```diff
@@ -0,0 +1,107 @@
...
+    input_objects.extend(samples)
+    resolwe = get_resolwe(*input_objects)
+
+    relations = resolwe.relation.filter(
```
dblenkus: This query must be further filtered. Imagine the case where `relation_filter` is empty (i.e. `cuffdiff` is called on a list of samples), this will pull **all** `case-control` relations that exists on the server.

resdk/tests/functional/analysis/e2e_differential_expressions.py
```diff
@@ -0,0 +1,35 @@
...
+    def test_cuffdiff(self):
+        collection = self.res.collection.create(name='Test collection')
+
+        # pylint: disable=unbalanced-tuple-unpacking
+        cuffquant_1, cuffquant_2, cuffquant_3, cuffquant_4 = self.get_cuffquants(4, collection)
+        # pylint: enable=unbalanced-tuple-unpacking
+        gff = self.get_gtf(1)
+
+        relation = collection.create_compare_relation(
+            samples=[cuffquant_1.sample, cuffquant_2.sample, cuffquant_3.sample,
```
dblenkus: Write this as:

    samples=[
        cuffquant_1.sample,
        cuffquant_2.sample,
        cuffquant_3.sample,
        cuffquant_4.sample,
    ],

resdk/tests/functional/analysis/e2e_differential_expressions.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+
+        # Run cuffdiff on a collection
+        cuffdiff = collection.run_cuffdiff(annotation=gff)
+        self.assertEqual(cuffdiff.input['case'], [cuffquant_3.id, cuffquant_4.id])
+        self.assertEqual(cuffdiff.input['control'], [cuffquant_1.id, cuffquant_2.id])
+        self.assertEqual(cuffdiff.input['annotation'], gff.id)
+
+        # Run cuffdiff on a relation
+        cuffdiff = differential_expressions.cuffdiff(relation, annotation=gff)
```
dblenkus: You can call this as `relation.run_cuffdiff(annotation=gff)`

resdk/tests/functional/analysis/e2e_differential_expressions.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TestExpressions(BaseResdkFunctionalTest):
+
+    def test_cuffdiff(self):
+        collection = self.res.collection.create(name='Test collection')
+
+        # pylint: disable=unbalanced-tuple-unpacking
+        cuffquant_1, cuffquant_2, cuffquant_3, cuffquant_4 = self.get_cuffquants(4, collection)
+        # pylint: enable=unbalanced-tuple-unpacking
+        gff = self.get_gtf(1)
```
dblenkus: This `1` is id of collection to which gtf objects will be added. You probably don't want to do this.

Use either `self.get_gtf(collection.id)` or `self.get_gtf()`.

resdk/tests/functional/analysis/e2e_differential_expressions.py
```diff
@@ -0,0 +1,35 @@
+# pylint: disable=missing-docstring
```
dblenkus: Write some more tests that covers edge cases. I.e. what happens if some samples from the collections are not in any relation,...

resdk/analysis/differential_expressions.py
```diff
@@ -0,0 +1,107 @@
...
+        label='case-control',  # this might not be needed
+        **relation_filter
+    )
+
+    for relation in relations:
+        control = []
+        case = []
+
+        sample_ids_rel = [sample.id for sample in get_samples(relation)]
+        sample_ids_res = [sample.id for sample in samples]
```
dblenkus: You can move this line out of for as it doesn't affect it.

resdk/analysis/differential_expressions.py
jvrakor: This query will pull all case-control relations that are in any way related to the samples (i.e. any sample from the resource is in this relation). This is further connected to skipping unsuitable relations mentioned later. But most important: all `case-control` relations from the server will NOT be pulled.

jvrakor: This is connected to skipping unsuitable relations mentioned later. That is also the fail-safe. We can of course decide on a different implementation.

resdk/analysis/differential_expressions.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+
+    for relation in relations:
+        control = []
+        case = []
+
+        sample_ids_rel = [sample.id for sample in get_samples(relation)]
+
+        if not set(sample_ids_rel).issubset(sample_ids_res):  # optional, not yet decided
+            continue
```
dblenkus: I think it would be better to skip (in next `for`) just samples that are not in `samples`.

resdk/analysis/differential_expressions.py
```diff
@@ -0,0 +1,127 @@
...
+        inputs['library_normalization'] = library_normalization
+
+    if dispersion_method is not None:
+        inputs['dispersion_method'] = dispersion_method
+
+    if threads is not None:
+        inputs['threads'] = threads
+
+    samples = get_samples(resource)
+    sample_ids_res = [sample.id for sample in samples]
```
dblenkus: Rename this to `sample_ids`.

resdk/analysis/differential_expressions.py
```diff
@@ -0,0 +1,127 @@
...
+        **relation_filter
+    )
+
+    cuffdiff_objects = []
+    for relation in relations:
+        control = []
+        case = []
+        for sample, position in zip(relation.samples, relation.positions):
+            if sample.id not in sample_ids_res:
+                continue
```
dblenkus: Add a new line after `continue`.

resdk/tests/functional/analysis/e2e_differential_expressions.py
```diff
@@ -0,0 +1,91 @@
...
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+from resdk.analysis import differential_expressions
+from resdk.tests.functional.base import BaseResdkFunctionalTest
+
+
+class TestExpressions(BaseResdkFunctionalTest):
+
+    def test_cuffdiff(self):
+        collection = self.res.collection.create(name='Test collection')
```
dblenkus: Rename these to `collection_1` and `collection_2`.

mstajdohar:
Write tests to improve coverage.

dblenkus:
There are nice e2e tests for added functionality. Problem is that they are run on Jenkins, but coverage report is not pushed to codecov.

@tjanez can you fix this?

PR #100, opened 2017-04-10
Support permissions management

dblenkus:
TODO:
- [x] tests

PR #119, opened 2017-06-02
Rafactor docs structure and add tutorials

mstajdohar:


jvrakor:
Suggestions by @jinlajf from July 13

>>> 1. It'd be great to link to the intro tutorial at the beginning (before the first command) - so that people can get a quick refresher as to how to link to the server ... 

2. Can we have link to mouse genome open in new tab by default? 

3. "or data whose sample is not yet inside..." -> "or data objects that have not yet been linked to a sample, or the sample included into the collection."

4. After "where instead of <First and Last Name> you insert your name." insert another sentence: "We ask you to follow this naming convention so that we have an easier job managing data on the server."

5. This command is a bit unusual because of the asterisk ("collection.add_samples(*samples)"). Do you think you can explain it in a simple sentence? If not, leave it. 

6. You choose simple variable names (eg. sample, collection) which is fine. The potential confusion can arise since most of these words are "special words" in ReSDK, e.g. genome is also the parameter name in the run_hisat2 method ... I know it's more verbose, but perhaps it would make sense to make the variable names just a tad less generic, e.g. genome -> ref_genome or something. Please consult with Miha. 

7. Could you include just a single-sentence figure caption with every figure, and/or at least reference the figure in the main text? I think that would help a lot. 

8. The downloaded yml relations file has some awquard wordings, e.g. "Below are listed all samples from this collections". I can go through this in more detail later, just something to be aware of. 

9. "We are now ready to run some expressions and differential expressions analysis." -> "We are now ready to run some expression and differential expression analyses."

10. command
# Query cuffnorm data object
collection.data.filter(type='data:expressionset:cuffnorm')
is probably not necessary, as we got exactly the same output with the previous command (run_cuffnorm()). No real harm in repeating it, though, at least we learn about a new type of data. 

11. Output on the status of process is not selfexplanatory: 
In [38]: cuffdiff.status
Out[38]: u'PR'
Since this is a ReSDK thing, could you explain a bit what each of possible statuses is?

jvrakor:
Suggestions from @jinlajf  from June 28

> - refer to the SDK consistently as ReSDK - check with Miha or Domen for correct capitalization of the acronym. 
> 
> - potential problem: what if pip not installed? I remember most of BioBash workshop participants had this problem. 
> 
> - introduction into what Python interpreter is is really misplaced - anyone who'll be reading this probably knows what a Python interpreter is, they have to know what CL is etc. I guess you just remove the big description and deal with that with a half sentence, e.g. "... start Python interpreter by typing 'python' into the command line, you'll recognize the interpreter by '>>>'" I know Rafe suggested that, I overrule. :) 
> 
> - A crucial snippet of code is missing (the one from example_index.py - how to connect to server) - I found the file and could log in, but others won't have your repository at their fingertips. :)  
> 
> - Please add a more helpful comment as to what start_logging() function does (why it's used here). 
> 
> - res.data.all() - gives an UnicodeEncodeError, not sure why .. 
> 
> - "We now have genome and reads data objects. We can continue by inspecting both of these objects to find out who created them"
> No need to copy the same comand for both objects. Rather tell the user they can learn about the object by calling certain attributes, e.g. 'output' or 'contributor', and give one example each. (If people used ipython, you could also instruct them to use the . and auto-complete after the object name to learn about other possible attributes. This is complicated to explain, though, and I guess most users will figure this out for themselves ...)
> 
> - "Often the next step the bioinformatician will do is align raw reads to a genome. It will result in a bam alignment file, bai index, unmapped reads and stats file."
> This is technically not true. Many aligners will output unmapped reads only if instructed to do so, and stats is often only printed to command line (=not a file by default). And .bai index is IMO never created automatically, but needs to be created separately with SAMTOOLS package. 
> So you could say something like: "A frequent procedure in bioinformatics is to is align sequencing reads to a reference genome. This is done by running a certain /process/. A process uses an algorithm or a sequence of algorithms to turn given inputs into outputs. Here we will use HISAT2 alignment process, but many more processes are available (see Process catalog). This process automatically creates a BAM alignment file, BAI index, a file with unmapped reads, and a stats file."
> 
> - - I only get these attributes: 
> In [32]: bam.
> bam.append   bam.extend   bam.insert   bam.remove   bam.sort     
> bam.count    bam.index    bam.pop      bam.reverse  
> What's wrong? So I get errors when trying to execute the command below. 
> 
> In [23]: bam.update()
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> <ipython-input-23-d0aa067338f3> in <module>()
> ----> 1 bam.update()
> 
> AttributeError: 'list' object has no attribute 'update'
> 
> 
> In [32]: bam.status
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> <ipython-input-32-98e5ee756fda> in <module>()
> ----> 1 bam.status
> 
> AttributeError: 'list' object has no attribute 'status'

dblenkus:
> * potential problem: what if pip not installed? I remember most of BioBash workshop participants had this problem.

You can add `Installing prerequisites` as we have in [Resolwe documentation](http://resolwe.readthedocs.io/en/latest/contributing.html?highlight=pip#installing-prerequisites).

> * res.data.all() - gives an UnicodeEncodeError, not sure why ..

This is already fixed in #100.

>I only get these attributes:
In [32]: bam.
bam.append bam.extend bam.insert bam.remove bam.sort
bam.count bam.index bam.pop bam.reverse
What's wrong? So I get errors when trying to execute the command below.

I will look at thus next week.

>This command is a bit unusual because of the asterisk ("collection.add_samples(*samples)"). Do you think you can explain it in a simple sentence? If not, leave it.

One option is to support the call without asterisk.

docs/tutorial-diffexp.rst
```diff
@@ -0,0 +1,319 @@
+.. _tutorial-diffexp:
+
+========================
+Differential expressions
+========================
+
+In this tutorial we will learn how to run a differential expression analysis. We will become familiar
+with the concepts of *sample* and *collection*. We will create a collection and add samples to that
```
jvrakor: What about "relations"?

My suggestion:
"...concepts of *sample*, *collection* and *relation*. We will create a collection, add samples to that collection and define relations between samples.

docs/tutorial-diffexp.rst
```diff
@@ -0,0 +1,319 @@
...
+	the collection.
+
+Now that we know what a collection is, let us create one:
+
+.. code-block:: python
+
+	collection = res.collection.create(name='Tutorial: <First and Last Name>')
+
+where instead of <First and Last Name> you insert your name. We ask you to follow this naming
+convention so that we have an easier job managing data on the server.We will now add queried
```
jvrakor: Whitespace missing: "...data on the server. We will now..."

docs/tutorial-diffexp.rst
```diff
@@ -0,0 +1,319 @@
...
+	removed from the server in a month.
+
+In the first tutorial (Getting Started) we ran HISAT2 aligment process on one of our samples.
+Now we want to run HISAT2 process on all of our samples at once. Since they are now all in the
+same collection, this can be easily accomplished by running HISAT2 on the whole collection:
+
+.. code-block:: python
+
+	# Get genome data object for the alignment
+	genome = res.data.get(1488)
```
jvrakor: Use `genome = res.data.get('genome-mm10-chr19')`

docs/tutorial-diffexp.rst
```diff
@@ -0,0 +1,319 @@
...
+Before we get to *cuffnorm* and *cuffdiff* analysis we will run *cuffquant*. This will save
+us some time later on because *cuffnorm* and *cuffdiff* will recieve quantified inputs insted of
+BAM files. For any of these analysis we will also need an appropriate annotation file (GTF, GFF).
+We can query and get it:
+
+.. code-block:: python
+
+	res.data.filter(type='data:annotation:gtf', created__date='2017-06-06')
+
+	annotation = res.data.get(1489)
```
jvrakor: Use `annotation = res.data.get('annotation-mm10-chr19')`

docs/start.rst
```diff
@@ -0,0 +1,223 @@
...
+
+	# Get data object by id
+	genome = res.data.get(1488)
+
+or
+
+.. code-block:: python
+
+	# Get data object by slug
+	genome = res.data.get('genome-mm10_chr19fastagz')
```
jvrakor: Use `genome = res.data.get('genome-mm10-chr19')`

mstajdohar:
@jvrakor I have addressed all your changes. Please do another review pass.

PR #120, opened 2017-06-07
Add running some helper functions on data objects

jvrakor:
@dblenkus @mstajdohar @jkokosar 

Add data as a resource for bowtie2 and hisat2 helper functions.

### DIFF
--- a/resdk/resources/utils.py
+++ b/resdk/resources/utils.py
@@ -155,11 +155,19 @@ def get_samples(resource):
     * call ResolweQuery object named `samples` (if exists) and return
       the result
     """
+    error_msg = ("Resource should be sample, have `samples` query, be list of multiple "
+                 "resources or be data object with not empty `sample` property.")
     if isinstance(resource, list):
         samples = []
         for res in resource:
             samples.extend(get_samples(res))
         return samples
+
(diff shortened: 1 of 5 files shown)


resdk/resources/utils.py
```diff
@@ -161,6 +161,14 @@ def get_samples(resource):
             samples.extend(get_samples(res))
         return samples
 
+    elif is_data(resource):
+        if not resource.sample:
+            raise TypeError(
+                "Resource should be sample, have `samples` query or be list of multiple resources."
```
dblenkus: You should add this option to the error message - i.e. Data object with not empty `sample` property.

Also, as the same message is used few lines below, save it to the variable and use it instead.

jvrakor: Also add some additional tests with data objects as inputs?

resdk/resources/utils.py
```diff
@@ -155,21 +155,28 @@ def get_samples(resource):
     * call ResolweQuery object named `samples` (if exists) and return
       the result
     """
+    error = ("Resource should be sample, have `samples` query, be list of multiple "
```
dblenkus: rename this to `error_msg`

PR #123, opened 2017-07-14
Return single object in analysis functions where applicable

dblenkus:
TODO:

- [x] tests

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -18,6 +18,8 @@ Added
 Changed
 -------
 * Use values instead of abbreviations for genome sizes in chip_seq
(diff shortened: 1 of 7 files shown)


jvrakor:
We might also want to get a single object if we run alignment on a single sample, collection with only one sample (or relation with only one sample but that maybe doesn't make sense), etc..

mstajdohar:
> We might also want to get a single object if we run alignment on a single sample, collection with only one sample (or relation with only one sample but that maybe doesn't make sense), etc..

I my opinion this should be consistent.

dblenkus:
@mstajdohar this is ready for review.

PR #125, opened 2017-08-22
Fix delete functionality for non-boolean force parameter types.

mzganec:


dblenkus:
Please add a change log entry.

PR #133, opened 2017-10-30
Use get_primary_bam utility in macs helper function

jvrakor:
@dblenkus @jkokosar @mstajdohar

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -9,6 +9,10 @@ All notable changes to this project are documented in this file.
 Unreleased
 ==========
 
+Added
+-----
+* Add ``get_primary_bam`` utility function
+
 Changed
 -------
 * Change ``macs14`` helper function to work on unannotated samples
@@ -16,6 +20,7 @@ Changed
   helper and test functions
 * **BACKWARD INCOMPATIBLE:** Remove ``update_knowledge_base`` script
 * Update contributing, start, and diffexp tutorial docs
+* Support primary bam files in ``macs`` helper function
 
 Fixed
 -----
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -63,7 +63,7 @@ Fixed
 
 Changed
 -----
-* Remove ``threads`` parameter from ``cuffquant`` and ``cufnorm``
+* Remove ``threads`` parameter from ``cuffquant`` and ``cuffnorm``
   helper functions
 
 Fixed
--- a/resdk/resources/sample.py
+++ b/resdk/resources/sample.py
@@ -21,6 +21,22 @@ class SampleUtilsMixin(object):
         """Return ``bam`` object on the sample."""
         return self.data.get(type='data:alignment:bam:')
 
+    def get_primary_bam(self, fallback_to_bam=False):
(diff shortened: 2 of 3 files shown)


resdk/resources/sample.py
```diff
@@ -19,7 +19,10 @@ def get_reads(self):
 
     def get_bam(self):
         """Return ``bam`` object on the sample."""
-        return self.data.get(type='data:alignment:bam:')
+        try:
+            return self.data.get(type='data:alignment:bam:primary:')
```
dblenkus: I don't like that `get_bam` function returns primary bam. You should add `get_primary_bam` function to make it cleaner and it can returns normal bam if primary is missing (you can also add flag to enable/disable this behaviour).

docs/CHANGELOG.rst
```diff
@@ -13,6 +13,8 @@ Changed
 -----
 * Remove ``threads`` parameter from ``cuffquant`` and ``cufnorm``
   helper functions
+* Modify ``get_bam`` utility function to correctly work with 
+  ``bam-split`` processed samples
```
dblenkus: Point out that this change is required to work with the latest process version - it is not a bug.

resdk/resources/sample.py
jvrakor: What about just adding an optional `primary` flag/input to `get_bam` function? Something like:
```def get_bam(self, primary=False):
       if primary:
           try:
               return self.data.get(type='data:alignment:bam:primary:')
           ...
       else:
           return self.data.get(type='data:alignment:bam:')```

File: docs/CHANGELOG.rst [hunk 3]
```diff
@@ -9,6 +9,15 @@ All notable changes to this project are documented in this file.
 Unreleased
 ==========
 
+Added
+-----
+* Add ``get_primary_bam`` utility function
+
+Changed
+-------
+* Replace `get_bam`` with ``get_primary_bam`` utility function in
```
dblenkus: Describe how this affects users, not how you achieve it, e.g. "Use only primary bam files in ``macs`` helper function"

resdk/resources/sample.py
```diff
@@ -21,6 +21,21 @@ def get_bam(self):
...
+
+        If the ``primary bam`` object is not present the function will
+        try to return a ``bam`` object if bam=True.
+
+        """
+        try:
+            return self.data.get(type='data:alignment:bam:primary:')
+        except LookupError:
+            if bam:
+                return self.get_bam()
```
dblenkus: I don't like this as ``get_primary_bam`` name implies that primary bam will be returned.

resdk/resources/sample.py
```diff
@@ -21,6 +21,21 @@ def get_bam(self):
         """Return ``bam`` object on the sample."""
         return self.data.get(type='data:alignment:bam:')
 
+    def get_primary_bam(self, fallback_to_bam=False):
+        """Return ``primary bam`` object on the sample.
+
+        If the ``primary bam`` object is not present the function will
+        try to return a ``bam`` object if fallback_to_bam=True.
```
dblenkus: Reword this to: 

    If the ``primary bam`` object is not present and ``fallback_to_bam`` is set
    to ``True``, a ``bam`` object will be returned.

PR #147, opened 2018-03-12
Add bamsplit helper function

jvrakor:
@dblenkus

### DIFF
--- a/resdk/analysis/chip_seq.py
+++ b/resdk/analysis/chip_seq.py
@@ -5,7 +5,50 @@ from resdk.resources.utils import (
     get_data_id, get_resource_collection, get_samples, is_background, is_sample,
 )
 
-__all__ = ('macs', 'rose2')
+__all__ = ('bamsplit', 'macs', 'rose2')
+
+
+def bamsplit(resource, header=None, header2=None):
+    """Run ``Bam split`` process on the resource.
+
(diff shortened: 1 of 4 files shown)


resdk/analysis/chip_seq.py
```diff
@@ -17,6 +17,48 @@ def is_background(sample):
...
+
+    results = []
+
+    if not isinstance(resource, list):
+        resource = [resource]
+
+    for single_resource in resource:
+        for sample in get_samples(single_resource):
+            bam = sample.get_bam()
+            inputs['bam'] = bam.id
```
dblenkus: Doesn't this work without `.id`? It think you can just write: `inputs['bam'] = sample.get_bam()`.

jvrakor: It doesn't work without `.id`. But I can rewrite it as  `inputs['bam'] = sample.get_bam().id`

dblenkus: It should work without `.id`. Please check why it doesn't. https://github.com/genialis/resolwe-bio-py/blob/master/resdk/resolwe.py#L292

mstajdohar: @jvrakor Would you print `type(sample.get_bam()).__name__`? The `get_data_id` function checks if `type(data).__name__ == 'Data'`, so my hypothesis is this condition is not satisfied.

PR #149, opened 2018-04-06
Add batch annotation function for collections

tristanbrown:
@jkokosar @jvrakor @dblenkus


resdk/data_upload/samplesheet.py
```diff
@@ -0,0 +1,263 @@
...
+                self.sample_annotation['sample']['optional_char'].append(
+                    '{0}:{1}'.format(option, entry[option]))
+
+    def validate(self):
+        """Validate the annotation spreadsheet file."""
+        # Check column headers
+        diff1 = COLUMNS - set(self.entry.keys())
+        diff2 = set(self.entry.keys()) - COLUMNS
+        err_msg = (
+            "Headers `{0}` {1}. You should use the"
```
dblenkus: Use `{}` instead of `{<position>}` if you only use each argument once.

dblenkus: And use `'` instead of `` ` ``.

resdk/data_upload/samplesheet.py
```diff
@@ -0,0 +1,263 @@
...
+            " headers from Annotation_spreadsheet.xlsm found on "
+            "https://github.com/genialis/resolwe-bio-py/tree/master"
```
dblenkus: This doesn't support versioning. I don't like the idea of downloading templates from github at all.

resdk/data_upload/samplesheet.py
```diff
@@ -0,0 +1,263 @@
...
+            "/resdk/data_upload."
+        )
+        if diff1:
+            raise KeyError(
+                err_msg.format(', '.join(diff1), "are missing")
```
dblenkus: `', '` -> `"', '"`, so each header will be in separate quotes.

resdk/tests/functional/base.py
```diff
@@ -87,7 +96,7 @@ def get_reads(self, count=1, collection=None):
             collections.
         :type collection: None, int or `~resdk.resources.Collection`
         """
-        reads_path = os.path.join(FILES_PATH, 'dummy_reads.fastq')
```
dblenkus: Why have you renamed this?

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,95 @@
...
+        sample_5 = reads_5.sample
+        sample_5.name = 'missing source'
+        sample_5.save()
+
+        # Apply the sample annotations from a local spreadsheet
+        samplesheet = self.get_samplesheet()
+        collection.annotate(samplesheet)
+
+        # Annotations from the example sheet for Samples 1, 2, and 3
+            # pylint: disable=line-too-long
```
dblenkus: Make lines shorter.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,95 @@
...
+        ann_1 = {'sample': {
+            'genotype': 'ANIMAL 1:\xa0PBCAG-FUS1, PBCAG-eGFP, PBCAG-mCherry, GLAST-PBase,\xa0PX330-P53',
+            'cell_type': 'Mixed',
+            'optional_char': ['AGE:38 days', 'LIBRARY_STRATEGY:Illumina Standard Prep ', 'OTHER_CHAR_1:2x75 bp', 'OTHER_CHAR_2:subdural cortical tumor, frontal/lateral location. Easily isolated sample', 'TISSUE:Tumor'],
+            'strain': 'N/A',
+            'source': 'Tumor',
+            'organism': 'Rattus norvegicus',
+            'molecule': 'total RNA',
+            'annotator': 'Tristan Brown',
+            }}
```
dblenkus: Wrong indentation.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,95 @@
...
+        sample_5.name = 'missing source'
+        sample_5.save()
+
+        # Apply the sample annotations from a local spreadsheet
+        samplesheet = self.get_samplesheet()
+        collection.annotate(samplesheet)
+
+        # Annotations from the example sheet for Samples 1, 2, and 3
+            # pylint: disable=line-too-long
+        ann_1 = {'sample': {
```
dblenkus: New line before `'sample'`

resdk/tests/functional/base.py
```diff
@@ -79,6 +79,15 @@ def get_gtf(self, collection=None, species='Homo sapiens',
             **collections
         )
 
+    def get_samplesheet(self):
```
dblenkus: Move this to the actual test case as this is not a general file, but is useful only for specific example.

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,85 @@
...
+        try:
+            # Find the existing (unannotated) samples
+            new_samples = collection.samples.filter(name=entry['SAMPLE_NAME'])
+            if not new_samples:
+                raise NameError(
+                    "Sample `{}` not found.".format(entry['SAMPLE_NAME'])
+                )
+
+            # Create and validate the sample descriptor
+            sample_info = Sample(entry)
```
dblenkus: Move this to `FileImporter`, so it returns a list of `Sample` objects.

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,85 @@
...
+
+            # Apply the sample descriptor to the sample(s)
+            for s in new_samples: # pylint: disable=C0103
```
dblenkus: Remove `pylint: disable=C0103`.

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,85 @@
...
+                s.descriptor = sample_info.sample_annotation
+                if sample_info.tag_community():
+                    s.tags.append(sample_info.tag_community())
+                s.save()
+                s.confirm_is_annotated()
+                sample_count += 1
+
+                # Apply the reads descriptor to all of the reads associated with
+                # the sample
+                for reads in s.data.filter(type='data:reads'):
```
dblenkus: `data:reads` -> `data:reads:`

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,85 @@
...
+                raise NameError(
+                    "Sample `{}` not found.".format(entry['SAMPLE_NAME'])
+                )
+
+            # Create and validate the sample descriptor
+            sample_info = Sample(entry)
+
+            # Apply the sample descriptor to the sample(s)
+            for s in new_samples: # pylint: disable=C0103
+                s.descriptor_schema = 'sample'
```
dblenkus: Is it ok to always use the same descriptor? We have many of them: https://github.com/genialis/resolwe-bio/tree/master/resolwe_bio/descriptors

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,85 @@
...
+    # else:
+    #     sheet = collection.data.get(name=samplesheet_name,
+    #                                 type='data:annotation')
+    #     sheet.download()
+    #     samplesheet = FileImporter(sheet.files()[0])
+    #     err_samples = []
+
+    # Apply each annotation to a sample in the collection
+    for entry in samplesheet.sample_list:
+        try:
```
dblenkus: Don't put too much code into `try`/`except` block. Only use it for lines that are expected to fail and handle specific errors.

.pylintrc
```diff
@@ -99,7 +99,7 @@ evaluation=10.0 - ((float(5 * error + warning + refactor + convention) / stateme
 [BASIC]
 
 # Good variable names which should always be accepted, separated by a comma
-good-names=i,j,k,ex,Run,_,logger
```
dblenkus: Remove this. You can use `ex` instead of `e`.

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,85 @@
+"""Annotate samples in a collection."""
+
+from __future__ import absolute_import, division, print_function, unicode_literals
+from resdk.data_upload.samplesheet import FileImporter, Sample
+
+def annotate_samples(collection, samplesheet_name, local_sheet=True,
+                     verbose=True):
+    """Annotate the samples in a collection based on a sample annotation
```
dblenkus: First line of docstring should be one-liner. Also add description of all parameters.

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,85 @@
...
+            if verbose:
+                print(e)
+
+    # Report annotation results and errors.
+    print("\nAnnotated {} samples.".format(sample_count))
+    if missing_samples:
+        print(
+            "\nThe following samples are missing from the collection:"
+            " {}. \nPlease check the names.".format(', '.join(missing_samples))
+            )
```
dblenkus: Wrong indentation.

resdk/data_upload/samplesheet.py
```diff
@@ -0,0 +1,263 @@
...
+        for var_name in [
+                'annotator',
+                'source',
+            ]:
+            var = self.entry[var_name.upper()]
+            if var.upper() in EMPTY:
+                raise ValueError(
+                    "For the sample, `{0},` `{1}` "
+                    "is not a valid {2}.".format(
+                        self.entry['SAMPLE_NAME'], var, var_name.upper())
```
dblenkus: Closing bracket into a new line.

setup.py
```diff
@@ -47,6 +47,7 @@
         'six>=1.10.0',
         'pyyaml>=3.11',
         'wrapt>=1.10.8',
+        'xlrd>=1.1.0'
```
dblenkus: Trailing comma.

dblenkus:
How is this different from https://github.com/genialis/resolwe-bio-py/pull/148/?

resdk/data_upload/samplesheet.py
tristanbrown: Would it be better if this were a data object hosted in a public collection on one of the production servers?

resdk/tests/functional/base.py
tristanbrown: `dummy_reads.fastq` was returning errors upon attempting upload, but reads.fastq.gz seems to work. Should this be a separate commit?

resdk/data_upload/annotate_samples.py
tristanbrown: I did it this way to give the NameError associated with missing samples a higher priority than the ValueError associated with invalid data. If the list of Sample objects is created first, then they must all be validated first (before being compared to the collection samples), and the ValueErrors will necessarily trigger before the NameErrors. 

One way around this would be to have `FileImporter` keep track of invalid samples without eliminating them altogether, and then let `annotate_samples` filter them out later, after checking for missing samples. The advantage here is that then both missing samples and invalid data could be reported for the same entries. I originally thought this would be a needlessly convoluted approach, but do you think it may be worth it?

tristanbrown:
I've switched to generating the annotation template using an export function, that also populates the samples from a collection. 
@dblenkus please re-review.

tristanbrown:
@dblenkus on second thought, it's not yet ready. I will first fix the problems causing the checks to fail.

tristanbrown:
@dblenkus Okay, all set.

dblenkus:
Overall this looks ok to me, @mzganec please check if this is conceptually ok and makes sense for general use.

tristanbrown:
Precluded by the first commit of #155.

PR #150, opened 2018-04-12
Fix get_bam and get_primary_bam functions

tristanbrown:
@jkokosar @dblenkus

### DIFF
--- a/resdk/query.py
+++ b/resdk/query.py
@@ -145,9 +145,9 @@ class ResolweQuery(object):
         # pylint: disable=protected-access
         if not isinstance(index, (slice,) + six.integer_types):
             raise TypeError
-        if ((not isinstance(index, slice) and index < 0) or
-                (isinstance(index, slice) and index.start is not None and index.start < 0) or
-                (isinstance(index, slice) and index.stop is not None and index.stop < 0)):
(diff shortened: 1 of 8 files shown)


resdk/query.py
```diff
@@ -145,9 +145,9 @@ def __getitem__(self, index):
         # pylint: disable=protected-access
         if not isinstance(index, (slice,) + six.integer_types):
             raise TypeError
-        if ((not isinstance(index, slice) and index < 0) or
```
dblenkus: Don't change the code, if the functionality stays the same.

resdk/resolwe.py
```diff
@@ -23,7 +23,7 @@
 import yaml
 # Needed because we mock requests in test_resolwe.py
 from requests.exceptions import ConnectionError  # pylint: disable=redefined-builtin
-from six.moves.urllib.parse import urljoin  # pylint: disable=import-error
```
dblenkus: Fix the problem instead of disabling the checker.

resdk/query.py
dblenkus: Ok, I see. You have to argue such changes better and put a description in the commit message.

tristanbrown:
Superseded by https://github.com/genialis/resolwe-bio-py/pull/151

PR #152, opened 2018-04-17
Change parameter email to username

JureZmrzlikar:


JureZmrzlikar:
IMO, this will **not** break many things. In essence, this PR only fixes the **misleading docstring**. As you may see, we already allow **only** username in `Resolwe` constructor.

The change is breaking "API" only for scripts `reads.py` and `sequp.py`. Still, they would not work if users provide their email, since login with email is not supported anyway...

dblenkus:
But this is still a production code that is used by clients, so we should do it better, especially as it can be achieved with a minimal amount of work.

resdk/query.py
```diff
@@ -204,6 +204,9 @@ def _clone(self):
     def _add_filter(self, filter_):
         """Add filter parameter."""
         for key, value in filter_.items():
+            # "sample" is called "entity" in the backend.
```
dblenkus: Use single quotes.

resdk/scripts/sequp.py
```diff
@@ -78,7 +78,7 @@ def sequp():
                                      'directory to the Resolwe server.')
 
     parser.add_argument('-a', '--address', help='Resolwe server address')
-    parser.add_argument('-e', '--email', help='User e-mail')
+    parser.add_argument('-e', '--username', help='User e-mail')
```
dblenkus: Update help message.

resdk/scripts/reads.py
```diff
@@ -38,7 +38,7 @@ def upload_reads():
 
     parser.add_argument('-a', '--address', default='https://torta.bcm.genialis.com',
                         help='Resolwe server address')
-    parser.add_argument('-e', '--email', default='admin', help='User name')
+    parser.add_argument('-e', '--username', default='admin', help='User name')
```
dblenkus: "User name" -> "Username"

PR #154, opened 2018-04-24
Make genome input work in cuffdiff helper function

jvrakor:
@dblenkus

### DIFF
--- a/resdk/analysis/differential_expressions.py
+++ b/resdk/analysis/differential_expressions.py
@@ -9,7 +9,7 @@ __all__ = ('cuffdiff', )
 
 
 def cuffdiff(resource, annotation, genome=None, multi_read_correct=None, fdr=None,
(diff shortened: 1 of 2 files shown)


resdk/analysis/differential_expressions.py
```diff
@@ -42,15 +42,14 @@ def cuffdiff(resource, annotation, genome=None, multi_read_correct=None, fdr=Non
...
-    :param int threads: use this many processor threads
 
     """
     inputs = {'annotation': get_data_id(annotation)}
 
     input_objects = [annotation]
 
     if genome is not None:
-        inputs['genome'] = genome
+        inputs['genome'] = genome.id
```
dblenkus: As already mentioned here https://github.com/genialis/resolwe-bio-py/pull/147#discussion_r179910866, this should work without `.id`. Please check why it doesn't.

PR #155, opened 2018-05-01
 Add batch upload helper functions for collections

tristanbrown:
upload_reads: Upload and annotate .fastq reads
upload_multi: Upload, demultiplex, and annotate .qseq reads

Relies on https://github.com/genialis/resolwe-bio-py/pull/149

@dblenkus @jkokosar @jvrakor


resdk/data_upload/reads.py
```diff
@@ -0,0 +1,108 @@
...
+    # Don't upload if the annotation info is invalid
+    if sample.name in pre_invalid:
+        raise ValueError(skip_msg + "Invalid annotation.")
+
+    # Check if filepaths were given
+    if not sample.path:
+        raise ValueError(skip_msg + "No forward reads given.")
+
+    # Check if at least one file has the proper filetype
+    validtypes = ['.fq', '.fastq']
```
dblenkus: This is duplicated form resolwe-bio. Fetch it from the api instead.

setup.py
```diff
@@ -47,6 +47,8 @@
         'six>=1.10.0',
         'pyyaml>=3.11',
         'wrapt>=1.10.8',
+        'xlrd>=1.1.0',
+        'openpyxl>=2.5.2',
```
dblenkus: Why do you need this?

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,195 @@
...
+        collection = self.res.collection.create(name='Test upload reads')
+
+        samplesheet = self.get_samplesheet()
+        collection.upload_reads(samplesheet)
+
+        # Check the uploaded data and created samples
+        self.assertEqual(len(collection.data), 2)
+        self.assertEqual(len(collection.samples), 2)
+        names = {sample.name for sample in collection.samples}
+        assert 'single-reads' in names
```
dblenkus: Use `self.assertIn`.

Jenkinsfile
```diff
@@ -141,7 +141,7 @@ throttle(["resolwe_bio_py"]) {
                                     // NOTE: End-to-End tests could hang unexpectedly and lock the
                                     // resource indefinitely thus we have to set a timeout on their
                                     // execution time.
-                                    timeout(time: 10, unit: "MINUTES") {
+                                    timeout(time: 15, unit: "MINUTES") {
```
dblenkus: Why do you need to change this?

resdk/analysis/__init__.py
```diff
@@ -1 +1,2 @@
 """Collection of utility functions for running analysis."""
+from . import _register
```
dblenkus: Why is this needed?

tristanbrown:
Jenkins is failing because it times out at 10 min. I have no idea why this is happening. On my machine, these commits add only 30-40 sec to the testing suite (which Jenkins spends 7-8 min on in the master branch).

dblenkus:
Please discuss all changes of testing environment with @tjanez before implementing them.

setup.py
tristanbrown: `xlrd` is used to import the annotation spreadsheets. `openpyxl` is used to export them.

resdk/analysis/__init__.py
tristanbrown: Originally, `resdk/__init__.py` patched the resources using:
```from .analysis import _register```

I wanted to similarly patch from `.data_uploads`, but this confused the namespace of `_register`. So, I instead had to patch using:
```
from . import analysis
from . import data_upload
```
and then include the `_register` imports in the sub-modules.

File: setup.py (hunk 2 above)
Comment (dblenkus) [reply setup.py]: Keep requirements to minimum. Find one stable and well maintained package and use it for all.

File: resdk/analysis/__init__.py (hunk 5 above)
Comment (dblenkus) [reply resdk/analysis/__init__.py]: This complicates things, you should probably do

    from .analysis import _register as analysis_register
    from .data_upload import _register as upload_register

and even importing just `_register`-s should probably work, only pylint will complain about.

File: resdk/data_upload/reads.py (hunk 1 above)
Comment (tristanbrown) [reply resdk/data_upload/reads.py]: Do we have a built-in function to do this? I can see how this information is stored using the `validate_regex` key on the various elements of `Process.input_schema`, but I'm not sure of the best way to parse this.

File: setup.py (hunk 2 above)
Comment (tristanbrown) [reply setup.py]: I guess `openpyxl` is the more important one, because it enables worksheet protection and restricted fields. So, I'll see if it will make sense to rewrite the import part based on that package.

Comment (tristanbrown) [reply setup.py]: I've eliminated the `xlrd` dependency.

File: resdk/data_upload/annotate_samples.py [hunk 6]
```diff
@@ -0,0 +1,116 @@
...
+    sample_count = 0
+
+    def _get_collection_samples(name):
+        """Check a collection for a given sample name.
+
+        :param str name: name of the sample(s) to look for
+        """
+        coll_samples = collection.samples.filter(name=name)
+        if not coll_samples:
+            missing_samples.add(name)
```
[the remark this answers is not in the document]
tristanbrown: I'm not sure what you mean by "map sample names to sample objects." Are you saying I should create a list of tuples of the type `(spreadsheet_sample, coll_sample)` using something like `map(lambda entry: (entry, collection.samples.filter(name=entry.name)))`, and then get the `missing_samples` list by filtering the mapping object for tuples containing an empty `coll_sample` field?

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+    if missing_samples:
+        logger.error(
+            "\nThe following samples are missing from the collection:"
+            " %s. \nPlease check the names.",
+            ', '.join(missing_samples)
+        )
+
+
+def export_annotation(collection=None, path=None):
```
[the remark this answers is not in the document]
tristanbrown: In the `_register.py` file, I have indeed patched `Collection` to include this method. I also included the `collection=None` default so that this function could still be used independently, perhaps to generate samplesheet templates that are not yet populated with samples. 

However, I can see how it might be useful to enforce this function only as a collection method. For example, then we could also eliminate the `path` variable and auto-generate a filename based on the collection name. This would encourage the standard of having one sample annotation spreadsheet belong to one collection, which will be useful in the future when we have a batch-annotation process accessible through the frontend. 

Do you think this would be desirable?

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,116 @@
...
+    ``Collection.annotate``.
+
+    :param str path: path to save the annotation spreadsheet template
+    """
+    if collection:
+        samples = collection.samples
+    else:
+        samples = []
+    samplesheet = FileExporter(path, samples)
+    samplesheet.export_template()
```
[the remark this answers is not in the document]
tristanbrown: `FileImporter` creates a data structure that is then explicitly accessed using `FileImporter.sample_list`, while `FileExporter` creates a data structure that is then explicitly written using `FileExporter.export_template`. So, it seems to me to already be analogous. But then is it still preferable to have `FileExporter` write a file immediately upon initialization?

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,116 @@
...
+    err_samples = samplesheet.invalid_samples
+    missing_samples = set()
+    sample_count = 0
+
+    def _get_collection_samples(name):
+        """Check a collection for a given sample name.
+
+        :param str name: name of the sample(s) to look for
+        """
+        coll_samples = collection.samples.filter(name=name)
```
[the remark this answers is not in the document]
tristanbrown: I've opened an issue for this here: https://github.com/genialis/meta/issues/298

tristanbrown:
I've substantially reorganized `annotate_samples.py` to address comments from @mzganec. I've requested further discussion on a few of the comments.

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,141 @@
...
+        logger.error(
+            "\nThe following samples are missing from the collection:"
+            " %s. \nPlease check the names.",
+            ', '.join(missing_names)
+        )
+
+    # Apply each annotation to a sample in the collection
+    logger.info("\nAnnotating Samples...\n")
+    sample_count = 0
+    for entry, samples in coll_samples:
```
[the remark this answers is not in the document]
tristanbrown: This does not work because `map` and `filter` objects are lazy-loaded in Python. The object needs to be evaluated in order to apply the functions or extract any values. If I try to filter the same object in several different ways, as soon as one is evaluated, it interferes with all of the other instances. 

If I try to use `map` to apply the annotations, they are not applied until the object is explicitly evaluated as a list. But because we don't actually need a list of return values from that operation, that method has poorer performance and uses more memory than simply using a `for` loop.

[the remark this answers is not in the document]
tristanbrown: In general, when no return values are required, `for` loops [actually give the best performance](http://leadsift.com/loop-map-list-comprehension/). When return values are required, list comprehension is usually better than `map`. 

Though, I suppose this performance data would be different if we were using numpy arrays.

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # Match annotation entries to collection samples
+    matched = map(
+        lambda entry: (entry, collection.samples.filter(name=entry.name)),
+        samplesheet.sample_list
+    )
+
+    # Partition out the samples missing from the collection
+    coll_samples, missing_samples = _part_list(lambda pair: pair[1], matched)
+    err_missing = [
```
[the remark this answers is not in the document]
tristanbrown: Because this function may be used by customers, the purpose of this is to ensure that all problems with their annotation attempt are reported on the first try. I don't suppose we want them to have to correct all of their annotations before they see that their sample isn't even present in the collection.

resdk/data_upload/annotate_samples.py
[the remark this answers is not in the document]
mzganec: I'm not concerned about performance of this `for` loop. Each iteration (sample annotation) takes orders of magnitude longer than the iteration time difference between the two implementations anyway. If we are concerned about memory footprint with big spreadsheets we should have implemented `FileImporter` as a generator in the first place.

I made this suggestion to improve code readability and possibly allow straightforward support for concurrency. But let's move on, @dblenkus will have the final say on this PR anyway.

resdk/data_upload/annotate_samples.py
```diff
@@ -0,0 +1,139 @@
...
+            "\nInvalid annotations were provided for the following samples: %s."
+            "\nPlease fill in all bolded columns of the template "
+            "generated by the `export_annotation` method of"
+            " your collection.",
+            ', '.join(err_samples)
+        )
+
+    # Match annotation entries to collection samples
+    matched = map(
+        lambda entry: (entry, collection.samples.filter(name=entry.name)),
```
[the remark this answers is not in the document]
tristanbrown: Because of the danger of overwriting all samples in a collection in the event of a `.filter(name='')` query, I'm just going to make it so that a sample annotation will not be applied if multiple samples are queried here.

tristanbrown:
@mzganec I've reorganized reads.py. I still need to apply similar changes to multiplexed.py.

tristanbrown:
@mzganec I've reorganized multiplexed.py, and abstracted the upload/annotate core from that and reads.py into base.py. I think it's all ready for another review.

tristanbrown:
@mzganec I've completed the revisions addressing your latest comments.

mstajdohar:
@mzganec I'd appreciate a fast review.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,268 @@
...
+        #     }
+        # }
+
+        ann_3 = {
+            'sample': {
+                'genotype': 'AX4',
+                'cell_type': '',
+                'optional_char': [
+                    'LIBRARY_STRATEGY:Illumina Standard Prep ',
+                    'OTHER_CHAR_1:300 bp'
```
mzganec: Add a trailing comma.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,268 @@
...
+        sample_1.name = 'Sample 1'
+        sample_1.save()
+
+        sample_2 = reads_2.sample
+        sample_2.name = 'Sample 2'
+        ann_2 = {
+            'sample': {
+                'genotype': '',
+                'cell_type':
+                'Mixed',
```
mzganec: Move to the line above.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,268 @@
...
+                'optional_char': [
+                    'LIBRARY_STRATEGY:Illumina Standard Prep ',
+                    'OTHER_CHAR_1:2x75 bp',
+                    'OTHER_CHAR_2:subdural cortical tumor, frontal/lateral'
+                    ' location. Easily isolated sample',
+                    'TISSUE:Tumor'
```
mzganec: Add a trailing comma.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,268 @@
...
+        }
+        sample_2.descriptor_schema = 'sample'
+        sample_2.descriptor = ann_2
+        sample_2.save()
+
+        reads_ann = {
+            'experiment_type': 'RNA-Seq',
+            'protocols': {
+                'growth_protocol': 'N/A',
+                'treatment_protocol': 'Control'
```
mzganec: Add a trailing comma.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,268 @@
...
+            }
+        }
+        reads_2.descriptor_schema = 'reads'
+        reads_2.descriptor = reads_ann
+        reads_2.save()
+
+        # Export the new template
+        filepath = 'template1.xlsm'
+        try:
+            os.remove(filepath)
```
mzganec: Unnecessary.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,268 @@
...
+        except OSError:
+            pass
+
+        collection.export_annotation(filepath)
+        assert os.path.exists(filepath)
+        # TODO: Include a file hash check
+        os.remove(filepath)
```
mzganec: Isn't this file automatically deleted after the test is completed?

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,268 @@
...
+        sample_2b.update()
+        sample_3.update()
+
+        self.assertEqual(sample_1.descriptor, ann_1)
+        self.assertEqual(sample_1.data[0].descriptor, reads_ann_1)
+        self.assertEqual(sample_1.tags, ['community:rna-seq'])
+        self.assertEqual(sample_2.descriptor, ann_2)
+        self.assertEqual(sample_2b.descriptor, ann_2)
+        self.assertEqual(sample_3.descriptor, ann_3)
+        self.assertEqual(sample_4.descriptor, {})
```
mzganec: We should assert the logger message that the annotation is invalid as well.

resdk/tests/functional/data_upload/e2e_upload.py
```diff
@@ -0,0 +1,268 @@
...
+
+        # Check the uploaded data and created samples
+        self.assertEqual(len(collection.data), 2)
+        self.assertEqual(len(collection.samples), 2)
+        names = {sample.name for sample in collection.samples}
+        self.assertIn('single-reads', names)
+        self.assertIn('paired-reads', names)
+        sample1 = collection.samples.get(name='single-reads')
+        sample2 = collection.samples.get(name='paired-reads')
+        wait_process_complete(sample1.data[0], 1)
```
mzganec: Aren't tests waiting until processes are completed?

dblenkus: There no need to wait for processes to be completed. We are also considering disabling processes from being run in e2e tests.

In ReSDK tests you should just check that inputs were saved correctly on the server. From there on you can rely on resolwe-bio tests that object will be processed correctly.

resdk/data_upload/multiplexed.py
```diff
@@ -0,0 +1,182 @@
...
+    elif path and not path2:
+        logger.debug('Uploading multiplexed data: %s', path)
+        qseq_reads = res.run('upload-multiplexed-single',
+                             input={
+                                 'reads': os.path.join(basedir, path),
+                                 'barcodes': os.path.join(basedir, path3),
+                                 'annotation': mapfile
+                             },
+                             collections=[collection])
+
```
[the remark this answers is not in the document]
tristanbrown: This is handled already in `_validate_multiplexed`.

resdk/data_upload/reads.py
```diff
@@ -0,0 +1,113 @@
...
+    if not any(ext in sample.path for ext in validtypes):
```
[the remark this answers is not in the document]
tristanbrown: The problem here is that the file extension checking in our upload processes in resolwe-bio is currently too permissive. For example, the `upload-fastq-single` process allows any `.bz2` file through, causing problems in my tests involving `.qseq.bz2` multiplexed files. If resolwe-bio would handle this case, then I could just catch the `ResolweServerError` here.

resdk/data_upload/samplesheet.py
```diff
@@ -0,0 +1,518 @@
...
+                "The sample name '%s' is duplicated. Please use unique "
+                "sample names.",
+                name
+            )
+            invalid_names.add(name)
+            valid_samples.pop(name, None)
+        else:
+            try:
+                valid_samples[name] = Sample(entry)
+            except ValueError as ex:
```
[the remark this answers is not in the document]
tristanbrown: I don't think we want to catch the `KeyError`, as it would indicated that the column headers have changed, which means the entire annotation spreadsheet is problematic.

resdk/data_upload/utils.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def wait_process_complete(data, recheck_interval):
+    """Suspend until the data is finished processing."""
+    busy = ['UP', 'RE', 'WT', 'PR']
+    while data.status in busy:
+        time.sleep(recheck_interval)
+        data.update()
+    if data.status == 'ER':
+        raise ValueError('Problem processing '
```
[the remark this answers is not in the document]
tristanbrown: This exception is caught in `_upload_multi_samples` to allow the rest of the samples to be processed. Keeping it as an exception ensures that `_create_multi_sample` doesn't try to create samples out of the outputs of this one.

resdk/tests/functional/data_upload/e2e_upload.py
tristanbrown: No. How would that happen with a file exported by ReSDK?

resdk/tests/functional/data_upload/e2e_upload.py
tristanbrown: I don't know how to to do this, and I can't find prior examples of it in ReSDK.

resdk/tests/functional/data_upload/e2e_upload.py
dblenkus: This is not a desired behaviour. If server's queue is filled up this would mean that script can run for a long time (e.g. days), which is not acceptable for users.

tristanbrown: In practice, these upload scripts already sometimes take days to run, simply because of the size and quantity of the files. So, we tend to run them on a dedicated server using screen. 

The way I have this iteration written now, the demultiplex processes will all be triggered first, and the sample creation will happen in bulk afterwards. So, if the script gets interrupted, the data will be on the servers, but the samples will not be created appropriately. 

One solution to this is I could turn the demultiplexed sample creation into a separate function call that can be independently triggered on any collection with multiplexed data objects. So then the whole pipeline can be triggered at once, or one step at a time (upload, recreate samples, annotate).

dblenkus: This is a really bad practice and once we will start using it there will be no way back.

The right way to handle this would be to make dss process which does everything. But this requires support for processes to fill sample (and data) descriptors. Until we support this, you should do the least ugly workaround, which in my opinion are independent functions.

tristanbrown:
I've reorganized the `collection.upload_multi` function a bit such that it no longer waits for the demultiplex process to complete prior to trying to recreate the samples. If the process is not complete for any data object at this point, the script will merely say so and continue on, treating the resulting samples as missing from the collection for the purposes of annotation. The user can then go back after the processes are complete and use `collection.resume_multi` to complete the sample-creation and annotation procedure after-the-fact, using the same annotation spreadsheet. 

So, in practical use, these scripts will no longer sit around waiting for any processes. Nevertheless, there still must be a little bit of waiting in the e2e tests, to ensure that the demultiplexed samples are being properly queried, recreated, and annotated.

jkokosar:
@dblenkus and @mzganec Please re-evaluate the latest changes introduced by @tristanbrown

tristanbrown:
@tjanez Please take a look at the failed tests that are blocking this PR. Jenkins is timing out at 10 min. It was already at 8-9 min prior to my commits.

tjanez:
> @tjanez Please take a look at the failed tests that are blocking this PR. Jenkins is timing out at 10 min. It was already at 8-9 min prior to my commits.

The [last 4 successful End-to-End tests of the `master` branch (builds 41, 43, 44, 45) took between 2min 55s and 3min and 43s](https://ci.genialis.com/job/genialis-github/job/resolwe-bio-py/job/master/). Therefore, I don't see a reason to bump the timeout limit.

tristanbrown:
@tjanez Oh, interesting. Well, on my local machine, these commits only add about 45 sec to the e2e testing suite. It should then be even faster on the servers here. Is there any way I can get more information out of Jenkins to troubleshoot what exactly is causing it to hang up for 10 min and time out?

tristanbrown:
@tjanez @dblenkus I've updated my local instances of resolwe, resolwe-bio, and genialis-base, and started running them with python 3.6 (as in https://github.com/genialis/resolwe-bio-py/commit/89507fb7be7fa05d94e2454e9973b349ab8b09ae), and now the tests get stuck locally as well. It seems that the demultiplex process cannot run. All data objects (for all of the tests) get stuck at "preparing input data".

tristanbrown:
@tjanez I've added an internal timeout to prevent my tests from going into an infinite loop while waiting for the demultiplex process. However, it looks like Jenkins is failing to connect to the servers after today's updates.

tjanez:
> However, it looks like Jenkins is failing to connect to the servers after today's updates.

That should be fixed now.

tristanbrown:
Apparently, `e2e_chip_seq.py` fails as well.

tristanbrown:
@tjanez @dblenkus @mzganec @mstajdohar I've found the other loop and given that one a timeout as well. Now the tests properly terminate with assert errors instead of timing out. 

The issue seems to be that the upload processes (either for uploading fastq's or uploading and demultiplexing qseq's) are failing to complete and generate their child data objects when Jenkins is running them. The uploaded data objects get stuck in 'WT' status, and their files and child data objects never appear on the server. I cannot do a full e2e test of my upload and annotate code if Jenkins will not allow the processes to complete. 

For the two failing tests:
- I could make `test_upload_reads` work by merely asserting the creation of the parent data object, without asserting the success of annotation. But maybe this is okay, since batch annotation is already being tested by `test_annotate`. Though, proper sample creation would go untested.
- This is a bigger problem for `test_upload_multiplexed`, because sample creation is non-trivial. The `resume_multi` function checks the status of the qseq data objects before attempting to create samples or annotate anything. It seems fairly problematic to leave this part entirely untested.

tristanbrown:
@mzganec @dblenkus @mstajdohar I've commented out the parts of the tests that cannot complete, and included a `TODO` statement regarding annotations through resolwe-bio processes. 

The tests now pass.

mstajdohar:
@tristanbrown Please address Matjaz's comments.

tristanbrown:
I fixed issue #3 and made the tests slightly more robust by including a `col.export_annotation` step at the end of `test_upload_reads`. I explored how to further improve this robustness by doing a file hash check on the exported excel templates, but found that .xls* files do not hash reliably (Every time a .xls* file with the same content is created, it has a different hash, and even a different file size. Even opening a .xls* file and closing it without saving will change the hash). 

I will address issues #1 and #2 tomorrow.

tristanbrown:
@mzganec Regarding issue 1, the intended behavior is that the same file should not be uploaded multiple times to the same collection via the script. The purpose of this is so an interrupted upload can be resumed using the same annotation spreadsheet without starting over completely. I've fixed this issue with a `collection.update()` step, and by switching the check of existing files to a check of existing data objects.

tristanbrown:
Regarding issue 2, this is identical to the behavior of the demultiplex process. I've fixed it in my code for the samples that I recreate, but fixing this for the samples that my functions don't touch (`Not_Matched_...` and `Bad_Quality_...`) should be part of a more general fix to the process itself. I've opened an issue for that: https://github.com/genialis/meta/issues/345.

tristanbrown:
I have learned how to assert log content, and have incorporated this into my tests. Hopefully, this makes them much more robust. 

@mzganec @mstajdohar I think this addresses the last of the comments.

PR #156, opened 2018-05-18
Make queries of empty values return empty instead of all objects

tristanbrown:
Fixes https://github.com/genialis/meta/issues/298.
@mzganec @dblenkus @jkokosar

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -9,6 +9,10 @@ All notable changes to this project are documented in this file.
 Unreleased
 ==========
 
+Fixed
+-----
+* Make ``<resource>.filter(**kwargs)`` return an empty query instead of all
+  resolwe objects if any of the ``kwargs`` is an empty value
 
 ==================
 4.0.0 - 2018-04-18
--- a/resdk/query.py
+++ b/resdk/query.py
@@ -163,12 +163,12 @@ class ResolweQuery(object):
 
             start = 0 if index.start is None else int(index.start)
             stop = 1000000 if index.stop is None else int(index.stop)  # default to something big
-            new_query._offset = start
-            new_query._limit = stop - start
+            new_query._offset = [start]
+            new_query._limit = [stop - start]
             return new_query
 
-        new_query._offset = self._offset + index if self._offset else index
-        new_query._limit = 1
+        new_query._offset = [self._offset + index if self._offset else index]
+        new_query._limit = [1]
 
         query_list = list(new_query)
         if not query_list:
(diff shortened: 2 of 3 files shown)


tristanbrown:
As it is, the software does not currently support queries of the type `.filter(description='')`. This type of query will return a fully unfiltered response. 

As far as I can tell, this seems to be a limitation of the Python `requests` module. So long as this limitation exists, it seems better for such queries to return no objects rather than all objects.

dblenkus:
It looks like the `requests` package handles it correctly:

    In [1]: import requests

    In [2]: r = requests.get('https://example.org', params={'foo': ''})

    In [3]: r.request.url
    Out[3]: 'https://example.org/?foo='

And it is also consistent with the API. The `/api/collection?name=` url returns all elements.
However, data endpoint (which was already migrated to the Elasticsearch on the backend) works differently - `/api/data?name=` returns empty list.

So in my opinion there is nothing to fix in the ReSDK.

JureZmrzlikar:
> So in my opinion there is nothing to fix in the ReSDK.

The root question here is: does Resowle handle it in the desired way or not? IMO search for empty name should return only the data objects with empty names, not the full list. This is something that you should debate with other developers of Resolwe. When you agree on this, we can solve the issue pointed out here.

tristanbrown:
A few months ago, was the default behavior of queries changed to make `.filter(name='foo')` equivalent to `.filter(name__startswith='foo')`? That could be the change that introduced this issue. How was that change implemented?

dblenkus:
This was implemented on the backend and only affects data endpoint for now: https://github.com/genialis/resolwe/commit/c81be28fa6f942d75c6e809737ae4d7bf4a66185

JureZmrzlikar:
@dblenkus A heard that the plan is to migrate all endpoints in the same way as `data`? Is this true and do we have a timeline for this?

dblenkus:
Yes, this is the plan, but we (probably) won't migrate all of them in the near feature. And also keep in mind that we will upgrade to [Django Rest Framework Filters](https://github.com/philipn/django-rest-framework-filters) version 1.0 when it will be released. So that may also be a chance to unify filters behaviour.

In general all filters should behave the same and they should be fixed/modified in Resolwe.

jkokosar:
@dblenkus Should this PR be closed? Will you fix things in Resolwe instead?

mstajdohar:
Yes let's close this. @kostko will move Sample and Collection search to Elasticsearch this month. After this change, Data, Sample and Collection will work as expected. Process and DescriptorSchema will remain inconsistent, but nobody makes this kind-of queries over them anyway. We usually just get them by slug, and we will consider removing other filters.

PR #158, opened 2018-07-17
Increase chunk size in Data.stdout method

JureZmrzlikar:


mzganec:
Locally I get the same error as displayed on Travis.

dblenkus:
#circularimport

dblenkus:
You should move the constant to a separate file.

JureZmrzlikar:
Which file do you suggest, @dblenkus? `resdk.__init__.py`?

dblenkus:
Something like `resdk/constants.py`.

mstajdohar:
@dblenkus Please do another review pass.

resdk/resources/data.py
```diff
@@ -307,13 +307,16 @@ def stdout(self):
         :rtype: string
 
         """
+        # Prevent circular imports:
+        from resdk.resolwe import CHUNK_SIZE
```
dblenkus: Move `CHUNK_SIZE` to `constants.py` (or similar), so there will be no circular import problems.

PR #160, opened 2018-07-25
Use slug when dehydrating `DescriptorSchema` or `Process` objects

JureZmrzlikar:


resdk/analysis/chip_seq.py
```diff
@@ -104,7 +104,7 @@ def macs(resource, use_background=True, p_value=None):
     return results
 
 
-def rose2(resource, use_background=True, genome='HG19', tss=None, stitch=None, beds=None):
```
dblenkus: Typo in commit message - `m` -> `n`.

JureZmrzlikar:
@mstajdohar , @dblenkus I've made some ugly tricks, but they will be temporary. Please re-review and comment. Tests pass.

JureZmrzlikar:
I still need to test the tutorial to confirm that things are not broken. We should automate this task in the future.

resdk/tests/unit/test_base.py
```diff
@@ -48,12 +49,18 @@ def test_fields(self):
         self.assertEqual(base_resource.fields(), ('writable', 'update_protected', 'read_only'))
 
     def test_dehydrate_resources(self):
-        obj = BaseResource(resolwe=self.resolwe_mock, id=1)
```
dblenkus: Is there a specific reason why have you changed this?

PR #161, opened 2018-08-10
Allow base dir of upload source files to be specified

tristanbrown:
@mzganec @mstajdohar Trying out the batch upload functions on Taco, I discovered that we typically don't have write access to the directories where the users put their reads files. So, I had to adjust how the file directories are specified. This PR allows the user to specify the base directory of the reads files, rather than assuming it is the directory containing the annotation spreadsheet. 

I've tested this commit on Taco in a real-world use-case, and the uploads work now.

### DIFF
--- a/resdk/data_upload/base.py
+++ b/resdk/data_upload/base.py
@@ -2,7 +2,6 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
 
 import logging
-import os
 
 from resdk.data_upload.samplesheet import FileImporter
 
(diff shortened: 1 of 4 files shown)


tristanbrown:
It looks like Jenkins is being held up by https://github.com/genialis/meta/issues/356.

mstajdohar:
[Here's the fix](https://github.com/genialis/resolwe-bio-py/pull/160). Jure will finish the code in the beginning of next week.

PR #162, opened 2018-08-13
Add option to enter your credentials interactively

JureZmrzlikar:


resdk/resolwe.py
```diff
@@ -90,9 +90,15 @@ def __init__(self, username=None, password=None, url=None):
 
         if username is None:
             username = os.environ.get('RESOLWE_API_USERNAME', None)
+            if username is None:
+                print('Username: ', end='')
+                username = input()
```
dblenkus: `username = input("Username: ")`

resdk/resolwe.py
```diff
@@ -90,9 +90,15 @@ def __init__(self, username=None, password=None, url=None):
...
 
         if password is None:
             password = os.environ.get('RESOLWE_API_PASSWORD', None)
+            if password is None:
+                print('Password: ', end='')
+                password = input()
```
dblenkus: `password = getpass.getpass("Password: ")`

mstajdohar:
The existing implementation logs you in as an anonymous user if you do not provide credentials. Some tests rely on that. If you force the interactive flow, how do we handle the anonymous use case?

JureZmrzlikar:
> The existing implementation logs you in as an anonymous user if you do not provide credentials. Some tests rely on that. If you force the interactive flow, how do we handle the anonymous use case?

@mstajdohar Good one. I don't know. I see two options, but hrere may be more (and better):

1. We could add `no_credentials=False` parameter and use `True` in cases where annonnyumus login is required.
2. If username and password are both empty strings, we just set them back to `None` and we get the same state as before.

What are the practices for other command line logins? Any suggestion is appreciated. @dblenkus , do you have an opinion on that?

dblenkus:
You can support a two stage login, e.g.:

    res = Resolwe('https://...')
    res.login()

The first line logs you in as anonymous user and then changes credentials in second line.

JureZmrzlikar:
@mstajdohar Thanks for noticing: I've added a note about `login()` method in "Getting started section".

resdk/resolwe.py
```diff
@@ -111,6 +112,13 @@ def __init__(self, username=None, password=None, url=None):
 
         self.logger = logging.getLogger(__name__)
 
+    def login(self):
+        """Login to Resolwe server with username and password."""
+        username = input('Username: ')
+        password = getpass.getpass('Password: ')
+        self.auth = ResAuth(username, password, self.url)
```
dblenkus: This code is duplicated.

Use this function also in `__init__`, you can change it to something like: `login(self, username=None, password=None, public=False)`.

mstajdohar: I would suggest `noninteractive=False` instead of `public=False`.

dblenkus: I prefer `public`, because you can logout with `res.login(public=True)`. Or maybe you can add `logout` method for this. 

But i am against using private variables (i.e. `noninteractive`) in public functions. A solution would be to add third function that is called from both places.

JureZmrzlikar:
@dblenkus @mstajdohar 
1. I've changed the code so nothing is duplicated
2. I've decided to use `interactive=True` instead of `noninteractive=False`. It juts seems much more readble.
Please comment or suggest improvements...

resdk/resolwe.py
JureZmrzlikar: Why do you consider `noninteractive` as a "private variable"? It is just a way to enter or not enter credentials...

JureZmrzlikar: To argue against public... What do you do if user enters public=True and also enters his username and password?

mstajdohar:
I don't like `public=True`, because for public login users should not call `.login(...)` at all. They should just create the resolwe objects, i.e. `Resolwe(url)`. If we include `public=True` they may get an impression they need to first create a `Resolwe` objects and then also call login.

mstajdohar:
@dblenkus Please review if this is now OK.

PR #164, opened 2018-08-23
Add data_usage method

JureZmrzlikar:


mstajdohar:
Approved. Rebase and merge.

resdk/tests/unit/test_resolwe.py
```diff
@@ -464,6 +464,32 @@ def test_good_response(self, resolwe_mock, requests_mock, os_mock, open_mock):
         # Why 6? 2 files in self.file_list, each downloads 3 chunks (defined in response mock)
 
 
+class TestDataUsage(unittest.TestCase):
+
+    @unittest.skipIf(six.PY2, "Skip in case of python2.x")
```
dblenkus: Why this doesn't work in py2?

resdk/tests/unit/test_resolwe.py
```diff
@@ -464,6 +464,32 @@ def test_good_response(self, resolwe_mock, requests_mock, os_mock, open_mock):
...
+    def test_data_usage(self):
```
dblenkus: I doubt it is worth having such test as it is testing a really small part of whole functionality. It would be much better to write an e2e test that actually makes request to backend.

mstajdohar:
Do we support sorting data usage on the server?

JureZmrzlikar:
@dblenkus I've added e2e tests and removed the silly unit test.
@mstajdohar  I've added support for sorting on the backend and also enable this in SDK. As an example check the `test_ordering` e2e test.

mstajdohar:
Let's merge this. Thanks

PR #171, opened 2018-09-28
Run e2e tests with code form tutorials

JureZmrzlikar:


mstajdohar:
The tutorial scripts should be well formatted so that they make sense if users download them as a whole. They should be in the tutorial folder.

Tests for tutorial scripts should be separate, they should execute the script (mock some parts if necessary) and validate results.

docs/tutorial-create.rst
```diff
@@ -0,0 +1,180 @@
...
+process) and to run alignment. You can check the full signature of the :obj:`run() <resdk.Resolwe.run>` method.
+
+Run workflows
+=============
+
+As typical analysis of data is performed by running a process that outputs some data. This data is fed as input into
+another proces that produces another set of outputs. This output is then again fed into another proces... Basically
+analysis is a sequence of processes. Sometimes, this sequence is so commonly used that one want to simplify it's
+execution. This can be done by using so called "workflow". Workflows are special processes: then run a stack of
+processes. On the outside, they look exactly the same as a normal process: They have a process slug, inputs, outputs...
```
amiamrita: On the outside, they look exactly the same as a normal process and have a process slug, inputs, outputs...

PR #174, opened 2018-10-10
Validate url in Resolwe constructor

JureZmrzlikar:


resdk/resolwe.py
```diff
@@ -94,6 +93,15 @@ def __init__(self, username=None, password=None, url=None):
 
         self.logger = logging.getLogger(__name__)
 
+    def _validate_url(self, url):
+        if not re.match(r'http[s]?://', url):
+            raise ValueError("Please fix url so it starts with https:// or http://.")
```
mstajdohar: I suggest: "Server url must start with http(s)://"

resdk/resolwe.py
```diff
@@ -94,6 +93,15 @@ def __init__(self, username=None, password=None, url=None):
...
+
+        try:
+            requests.get(url)
+        except requests.exceptions.ConnectionError:
+            raise ValueError("The following page does not exist: {}".format(url))
```
mstajdohar: I suggest: "The site can't be reached: {}"

resdk/resolwe.py
romunov: Do we really want to send credentials over http?

resdk/tests/unit/test_resolwe.py
```diff
@@ -91,6 +91,19 @@ def test_init(self, resolwe_mock, resauth_mock, slumber_mock, resolwe_api_mock,
         self.assertEqual(resolwe_querry_mock.call_count, 10)
         self.assertEqual(log_mock.getLogger.call_count, 1)
 
+    @patch('resdk.resolwe.requests.get')
+    def test_validate_url(self, requests_get_mock):
+        resolwe = MagicMock(spec=Resolwe)
+
+        message = 'Please fix url so it starts with https:// or http://.'
```
romunov: Don't forget to fix this as well.

resdk/resolwe.py
JureZmrzlikar: AFAIK `http` si just for localhost. All real servers have `https`.

PR #177, opened 2018-10-17
Minor fixes

JureZmrzlikar:


resdk/resources/data.py
```diff
@@ -257,7 +236,8 @@ def put_in_download_list(elm, fname):
         if field_name and not field_name.startswith('output.'):
             field_name = 'output.{}'.format(field_name)
 
-        for ann_field_name, ann in self.annotation.items():
+        flattened = flatten_field(self.output, self.process_output_schema, 'output')
```
mstajdohar: Oh that's why we "needed" the self.annotations... I am happy you removed them.

PR #178, opened 2018-10-30
Handle multiple fastq objects in Sample.get_reads method

JureZmrzlikar:


resdk/resources/sample.py
```diff
@@ -15,9 +15,26 @@ class SampleUtilsMixin(object):
...
+        filtered, subsampled...), return the last one. If any other of
+        the ``fastq`` objects is required, one can provide additional
+        ``filter`` arguments and limits search to one result.
+        """
+        reads = self.data.filter(
+            type='data:reads:fastq',
+            ordering='-id',
+            **filters,
+        )
+        # TODO: In future, implement method ``last()`` on ResolweQuery
```
mstajdohar: Change 
```
# and use it instead or ordering by ``id``.
```
to
```
# and return the ResolweQuery object.
```

resdk/resources/sample.py
```diff
@@ -15,9 +15,26 @@ class SampleUtilsMixin(object):
...
+
+        If there are multiple ``fastq`` objects in sample (trimmed,
+        filtered, subsampled...), return the last one. If any other of
+        the ``fastq`` objects is required, one can provide additional
+        ``filter`` arguments and limits search to one result.
+        """
+        reads = self.data.filter(
+            type='data:reads:fastq',
+            ordering='-id',
+            **filters,
```
mstajdohar: If user defines the `type` filter, we should override the type.

docs/CHANGELOG.rst
```diff
@@ -21,6 +21,7 @@ Added
 
 Fixed
 -----
+* Handle multiple ``fastq`` objects in ``Sample.get_reads`` method
```
mstajdohar: Make more descriptive. Explain the latest read object will be returned.

resdk/resources/sample.py
```diff
@@ -15,9 +15,26 @@ class SampleUtilsMixin(object):
...
+        )
+        # TODO: In future, implement method ``last()`` on ResolweQuery
+        # and use it instead or ordering by ``id``.
+
+        if not reads:
+            raise LookupError('Matching object does not exist.')
```
mstajdohar: I suggest something like `Reads not found on sample {}`

resdk/resources/sample.py
```diff
@@ -15,9 +15,26 @@ class SampleUtilsMixin(object):
 
     """
 
-    def get_reads(self):
-        """Return ``fastq`` object on the sample."""
-        return self.data.get(type='data:reads:fastq')
+    def get_reads(self, **filters):
+        """Return ``fastq`` object in sample.
```
mstajdohar: Change to `Return the latest ``fastq`` object in sample.`

mstajdohar:
Tests would be nice.

JureZmrzlikar:
@mstajdohar  please have another review pass...

PR #179, opened 2018-11-05
Move creation of ResolweQuery instances to _login method

JureZmrzlikar:


JureZmrzlikar:
I consider this "technical" fix, that does not directly affect the user. AFAIK we don't put such changes in the changelog, do we?

mstajdohar:
It's a fix that will affect users (`res.resource` vs `res.resource.all()`). We should log bug fixes.

PR #181, opened 2019-01-07
Small features

JureZmrzlikar:


resdk/resources/base.py
```diff
@@ -26,6 +26,7 @@ class BaseResource(object):
     endpoint = None
     query_endpoint = None
     query_method = 'GET'
+    full_search_keyword = None
```
dblenkus: Rename to `full_search_paramater` as this is query parameter in url.

JureZmrzlikar:
@dblenkus I've addressed you comments. Please make another review pass.

JureZmrzlikar:
@dblenkus Please have another review pass when you find some time.

resdk/resources/base.py
```diff
@@ -235,6 +232,20 @@ def contributor(self):
 
         return self._contributor
 
+    @property
+    def created(self):
+        """Creation time."""
+        if 'created' not in self._original_values:
+            raise ValueError('Instance does not have `created` attribute.')
```
dblenkus: Can you add this only to applicable classes?

resdk/resources/data.py
```diff
@@ -201,6 +199,20 @@ def descriptor_schema(self, dschema):
         # Save descriptor schema if already hydrated, otherwise it will be rerived in getter
         self._hydrated_descriptor_schema = dschema if is_descriptor_schema(dschema) else None
 
+    @property
+    def started(self):
+        """Start time."""
+        if 'started' not in self._original_values:
+            raise ValueError('Instance does not have `started` attribute.')
```
dblenkus: This is not needed.

JureZmrzlikar: Why not? `resdk.resources.Data` can be created locally and no `started` value will exist in `_original_values`. Do we return `None` or do we raise an error? Same goes for all: `created`, `modified`, `started`, `finished`.

resdk/resources/base.py
JureZmrzlikar: What do you mean? The only classes that inherit from this base class are those that have `created` and `modified` attributes...

resdk/resources/data.py
dblenkus: I see. The usual approach that we use for such cases is `if not self.id: raise`.

JureZmrzlikar:
@dblenkus Comments addressed.

resdk/analysis/prepare_geo.py
```diff
@@ -43,8 +43,7 @@ def prepare_geo_chipseq(resource, name=None):
 
     """
     reads = []
-    macs14 = []
-    relations = []
+    macs = []
```
dblenkus: Don't forget to add a change log.

JureZmrzlikar: This was just a "technical fix". The sinature of the function `prepare_geo_chipseq` did not change. So for the end user, there is no visible difference. AFAIK such changes don't belong to changelog...

dblenkus: Ok, it makes sense.

PR #184, opened 2019-03-07
Add delete_content parameter to method BaseCollection.delete

JureZmrzlikar:


resdk/resources/collection.py
```diff
@@ -163,6 +163,23 @@ def download(self, file_name=None, field_name=None, download_dir=None):
 
         self.resolwe._download_files(files, download_dir)  # pylint: disable=protected-access
 
+    def delete(self, force=False, delete_content=False):
+        """Delete the resource object from the server."""
```
dblenkus: Mention `delete_content` in the docstring. It would be also nice to describe `force` parameter, but I know it is also not documented in the base class.

resdk/query.py
```diff
@@ -309,17 +311,31 @@ def filter(self, **filters):
...
+
+        :param bool force: Do not trigger confirmation prompt. WARNING: Be
+            sure that you really know what you are doing as deleted objects
+            are not recoverable.
+        :param bool delete_content: Also delete all the objects that the
+            current object contains.
+
+        """
+        kwargs = {'force': True}
+        if delete_content and isinstance(self.resource, resdk.resources.collection.BaseCollection):
```
dblenkus: Raise an error if the object is not a collection.

JureZmrzlikar: Done.

PR #187, opened 2019-04-26
Upgrade packaging

mstajdohar:


.travis.yml
```diff
@@ -8,20 +8,14 @@ language: python
 # NOTE: Explicit Python versions make Travis job description more informative
 matrix:
   include:
-    - env: TOX_ENV=py27
-      python: "2.7"
-    - env: TOX_ENV=py34
-      python: "3.4"
-    - env: TOX_ENV=py35
```
dblenkus: This commit also removes support for Python 3.5, not only Python 2

setup.py
```diff
@@ -77,6 +77,7 @@
         'Programming Language :: Python',
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.6',
+        'Programming Language :: Python :: 3.7',
```
dblenkus: This should go into a separate commit - together with changes in tox file.

.travis.yml
mstajdohar: I have changed to commit message

PR #193, opened 2019-07-23
Remove unused code and sync with latest Resolwe

JureZmrzlikar:


JureZmrzlikar:
@dblenkus , @mstajdohar Please review. Tests are still WIP, but after some playing around basic operations seem to work. Please provide a general feedback, i plan to polish this in the following days.

resdk/tests/functional/permissions/e2e_permissions.py
```diff
@@ -1,6 +1,7 @@
 # pylint: disable=missing-docstring
 from resdk.exceptions import ResolweServerError
-from resdk.tests.functional.base import ADMIN_USERNAME, USER_USERNAME, BaseResdkFunctionalTest
```
mstajdohar: Should we change the above import to a relative import too?

docs/tutorial-basics.rst
```diff
@@ -77,8 +77,8 @@ one Sample. Two distinct Samples cannot contain the same Data object.
 **Collection** is a group of Samples. In addition to Samples and their Data,
 Collections may contain Data objects that store other analysis results. Example
 of this are differential expressions - they are done as combination of many
-Samples and cannot belong to only one Sample. Samples and Data objects may be
-in multiple Collections.
+Samples and cannot belong to only one Sample. Each Sample and Data object can
+only be one Collection.
```
mstajdohar: `be in`

JureZmrzlikar: :+1:

resdk/tests/functional/permissions/e2e_permissions.py
JureZmrzlikar: Yes. Will do.

resdk/resources/data.py
```diff
@@ -48,171 +45,132 @@ def __init__(self, resolwe, **model_data):
...
-        self.process_type = None
-        #: process name
-        self.process_name = None
-        #: data object's tags
-        self.tags = None
         #: indicate whether `descriptor` doesn't match `descriptor_schema` (is dirty)
         self.descriptor_dirty = None
-        #: process slug
-        self.process_slug = None
+        #: annotation data, with the form defined in descriptor_schema
```
dblenkus: Is there any reason for reordering this in this commit?

resdk/resources/base.py
```diff
@@ -260,3 +274,16 @@ def __repr__(self):
...
         )
+
+    def _resource_setter(self, payload, resource, field):
+        """Set ``resource`` with ``payload`` on ``field``."""
+        if isinstance(payload, resource):
+            setattr(self, field, payload)
+        elif isinstance(payload, dict):
+            setattr(self, field, resource(resolwe=self.resolwe, **payload))
+        elif isinstance(payload, int):
+            setattr(self, field, resource(resolwe=self.resolwe, id=payload))
```
dblenkus: Populate the resource object.

resdk/resources/data.py
```diff
@@ -48,171 +45,132 @@ def __init__(self, resolwe, **model_data):
...
-            return None
-
-        if self._hydrated_descriptor_schema is None:
-            if isinstance(self._descriptor_schema, int):
-                query_filters = {'id': self._descriptor_schema}
-            else:
-                query_filters = {'slug': self._descriptor_schema}
+        """Get sample."""
+        if self._original_values.get('entity', None):
+            self._sample = Sample(resolwe=self.resolwe, **self._original_values['entity'])
```
dblenkus: You can load data into a `Sample` object in setter method, there is no overhead in doing that.

resdk/resolwe.py
```diff
@@ -255,28 +255,20 @@ def run(self, slug=None, input={}, descriptor=None,  # pylint: disable=redefined
...
             data['descriptor'] = descriptor
-            data['descriptor_schema'] = descriptor_schema
+            data['descriptor_schema'] = {'slug': descriptor_schema}
+
+        if collection:
+            data['collection'] = {'id': get_collection_id(collection)}
 
-        if collections:
-            data['collections'] = collections
+        if data_name:
```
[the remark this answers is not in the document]
JureZmrzlikar: I thought it might be nice that order resembles the order of arguments. Should have put it into separate commit.

resdk/tests/functional/docs/e2e_docs.py
```diff
@@ -33,14 +34,14 @@ def tearDown(self):
         shutil.rmtree(self.tmpdir)
 
         if hasattr(self, 'reads'):
-            self.reads.sample.delete(force=True)
-            self.reads.delete(force=True)
+            self.reads.sample.delete(force=True)  # pylint: disable=no-member
```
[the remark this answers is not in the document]
JureZmrzlikar: Because `self.reads` etc. are defined in subclasses.

resdk/resources/base.py
```diff
@@ -72,20 +70,37 @@ def _dehydrate_resources(self, obj):
...
     def save(self):
         """Save resource to the server."""
         def field_changed(field_name):
             """Check if local field value is different from the server."""
-            return getattr(self, field_name) != self._original_values.get(field_name, None)
+            original_value = self._original_values.get(field_name, None)
+            current_value = getattr(self, field_name, None)
+
+            if isinstance(current_value, BaseResource) and original_value:
+                return current_value.id != original_value.get('id', None)
```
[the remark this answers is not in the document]
JureZmrzlikar: I would omit this check as it is really cumbersome to get Resource class from field name. Can't we just trust our users not to do such mistakes? :-)

resdk/resources/data.py
JureZmrzlikar: No, should be put into a separate one.

resdk/resources/data.py
JureZmrzlikar: Consider something like:
`d = res.data.get(id=1)` and than `d.sample`
In such cases attribute setting works like this::

    for field_name in self.fields():
        setattr(self, field_name, payload.get(field_name, None))

The issue is that data for `sample` attribute is stored in `entity` key in the payload. We have two options: (1) as done now or (2) by defining something like `attribute_mapping = {'sample': 'entity'}` as class attribute and then::

    for field_name in self.fields():
        field_name = self.attribute_mapping.get(field_name, field_name)
        setattr(self, field_name, payload.get(field_name, None))

Do you prefer the second option?

resdk/resources/base.py
JureZmrzlikar: Please check if you like the solution now.

resdk/resources/relation.py
```diff
@@ -71,21 +67,21 @@ def samples(self):
...
-        self._collection = get_collection_id(collection)
-        # Save collection if already hydrated, othervise it will be rerived in getter
-        self._hydrated_collection = collection if is_collection(collection) else None
+        self._resource_setter(payload, Collection, '_collection')
 
     def update(self):
         """Clear cache and update resource fields from the server."""
         self._samples = None
-        self._hydrated_collection = None
+        self._collection = None
```
dblenkus: You no longer need to clear cache for these objects as they are not lazy loaded anymore.

JureZmrzlikar: But they ares still cached, so it makes sense to clear cache at update.

dblenkus: If related object is included in object it will automatically update it. Setting it to none only makes sense if related object is retrieved in a separate request and you mark that it needs to be done again.

JureZmrzlikar: Agh, sorry, you/re right. Removed.

PR #194, opened 2019-07-30
Add duplicate method to collection, sample and data resource

hadalin:
This seems to be working, but it needs some polishing:

- Better output when when a resource is duplicated. Currently the whole duplicated object (dict) is returned
- Tests
- Changelog entry
- SDK Reference?

### DIFF
--- a/resdk/resources/collection.py
+++ b/resdk/resources/collection.py
@@ -185,3 +185,12 @@ class Collection(CollectionRelationsMixin, BaseCollection):
             self._relations = self.resolwe.relation.filter(collection=self.id)
 
         return self._relations
+
+    @assert_object_exists
+    def duplicate(self):
+        """Duplicate (make copy of) ``collection`` object.
+
+        :return: Duplicated collection
+        """
+        duplicated = self.api().duplicate.post({'ids': [self.id]})
+        return self.__class__(resolwe=self.resolwe, **duplicated[0])
--- a/resdk/resources/sample.py
+++ b/resdk/resources/sample.py
@@ -225,3 +225,18 @@ class Sample(SampleUtilsMixin, BaseCollection):
             self._is_background = len(background_relations) > 0  # pylint: disable=len-as-condition
 
         return self._is_background
+
+    @assert_object_exists
+    def duplicate(self, inherit_collection=False):
+        """Duplicate (make copy of) ``sample`` object.
+
+        :param inherit_collection: If ``True`` then duplicated samples
+            (and their data) will be added to collection of the original
+            sample.
+        :return: Duplicated sample
+        """
+        duplicated = self.api().duplicate.post({
+            'ids': [self.id],
+            'inherit_collection': inherit_collection
+        })
+        return self.__class__(resolwe=self.resolwe, **duplicated[0])
--- a/resdk/tests/functional/duplication/e2e_duplication.py
+++ b/resdk/tests/functional/duplication/e2e_duplication.py
@@ -1,0 +1,31 @@
+# pylint: disable=missing-docstring
+from resdk.exceptions import ResolweServerError
+from resdk.tests.functional.base import USER_USERNAME, BaseResdkFunctionalTest
+
+
+class TestDuplication(BaseResdkFunctionalTest):
+
+    def test_collection_duplication(self):
+        collection = self.res.collection.create(name='Test collection')
+        duplicate = collection.duplicate()
+        self.assertEqual(duplicate.name, 'Copy of Test collection')
+
+        duplicate.delete(force=True)
+        collection.delete(force=True)
+
+    def test_sample_duplication(self):
+        sample = self.res.sample.create(name='Test sample')
+        duplicate = sample.duplicate()
+        self.assertEqual(duplicate.name, 'Copy of Test sample')
+
+        duplicate.delete(force=True)
+        sample.delete(force=True)
+
+    def test_data_duplication(self):
+        data = self.res.run(slug='test-sleep-progress', input={'t': 1})
+        # Let's not wait for processing to finish and
+        # check the expected exception to be raised.
+        with self.assertRaisesRegex(ResolweServerError, 'done or error status to be duplicated'):
+            # Data's `duplicate` raises an exception if status
+            # of the object is not done or error.
+            data.duplicate()
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -5,6 +5,15 @@ Change Log
 All notable changes to this project are documented in this file.
 
 
+==========
+Unreleased
+==========
+
+Added
+-----
+- Add duplicate method to collection, sample and data resources
+
+
 ===================
 11.0.1 - 2019-08-19
 ===================
--- a/resdk/resources/data.py
+++ b/resdk/resources/data.py
@@ -307,3 +307,12 @@ class Data(BaseResolweResource):
                 output += chunk
 
         return output.decode("utf-8")
+
+    @assert_object_exists
+    def duplicate(self):
+        """Duplicate (make copy of) ``data`` object.
+
+        :return: Duplicated data object
+        """
+        duplicated = self.api().duplicate.post({'ids': [self.id]})
+        return self.__class__(resolwe=self.resolwe, **duplicated[0])


resdk/resources/collection.py
```diff
@@ -264,3 +264,11 @@ def remove_samples(self, *samples):
...
 
         self.samples.clear_cache()
+
+    def duplicate(self):
+        """Duplicate (make copy of) ``collection`` object."""
+        if self.id is None:
+            raise ValueError('Instance must be saved before making a duplicate.')
+
+        duplicated = self.resolwe.api.collection().duplicate.post({'ids': [self.id]})
+        return duplicated[0]
```
dblenkus: You need to return a `Collection` object, something like `self.__class__(resolwe=self.resolwe, **duplicated[0])`

dblenkus:
And don't forget to add a change log.

dblenkus:
This looks good!

Sample endpoint also support [`inherit_collection`](https://github.com/genialis/resolwe/blob/c401f8363e10307cb2c507a1c8ec733cc1599e1e/resolwe/flow/views/entity.py#L103) parameter. Can you also support that? And don't forget to handle errors that can be raised.

resdk/tests/functional/duplication/e2e_duplication.py
```diff
@@ -0,0 +1,20 @@
+# pylint: disable=missing-docstring
+from resdk.tests.functional.base import BaseResdkFunctionalTest
+
+
+class TestDuplication(BaseResdkFunctionalTest):
+
+    def test_collection_duplication(self):
```
hadalin: @dblenkus travis-ci doesn't seem to [run these tests](https://travis-ci.org/genialis/resolwe-bio-py/jobs/565663036#L406-L413). Is this expected?

resdk/resources/sample.py
```diff
@@ -313,3 +313,22 @@ def is_background(self):
             self._is_background = len(background_relations) > 0  # pylint: disable=len-as-condition
 
         return self._is_background
+
+    def duplicate(self, inherit_collections=False):
+        """Duplicate (make copy of) ``sample`` object.
+
+        :param inherit_collections: If ``True`` then duplicated
```
hadalin: Added `inherit_collections` parameter.

hadalin:
>And don't forget to handle errors that can be raised.

Exceptions seem to be handled with `handle_http_exception`. Should we leave exception handling to the end user, instead of doing it for them?

resdk/resources/sample.py
dblenkus: This was renamed to `inherit_collection` in latest resolwe.

dblenkus:
It's fine if an error is raised, we just want to make sure it includes relevant informations about the underlying error.

If api returns meaningful response, you should use it for a user-friendly exception. Otherwise we will handle this when we will improve error handling in general, but that will require quite some refactoring on the backend.

hadalin:
What do you have in mind? Creating `DuplicationError(Exception)` and packing the original exception in it?

resdk/tests/functional/duplication/e2e_duplication.py
hadalin: I think I found the problem. There no `__init__.py` file in the directory.

hadalin: Though is seems that not one functional test is actually tested.

hadalin: Oh, never mind, I think Jenkins runs them.

JureZmrzlikar: Yes, e2e tests are run by Jenkins.

resdk/tests/functional/duplication/e2e_duplication.py
```diff
@@ -0,0 +1,39 @@
...
+        collection = self.res.collection.create(name='Test collection')
+        duplicate = collection.duplicate()
+        self.assertEqual(duplicate.name, 'Copy of Test collection')
+
+    def test_sample_duplication(self):
+        sample = self.res.sample.create(name='Test sample')
+        duplicate = sample.duplicate()
+        self.assertEqual(duplicate.name, 'Copy of Test sample')
+
+    def test_sample_duplication_perms(self):
```
JureZmrzlikar: IMO, this test does not belong here. In resdk, we should only verify that Resolwe endpoints work with a minimal example (e.g. creating a duplicated data/sample/collection). All edge cases, permissions, etc... should be tested in Resolwe.

resdk/tests/functional/duplication/e2e_duplication.py
```diff
@@ -0,0 +1,39 @@
...
+        duplicate = sample.duplicate(inherit_collection=True)
+        self.assertEqual(len(collection.samples), 1)
+        self.assertEqual(collection.samples[0].name, 'Test sample')
+        self.assertEqual(duplicate.name, f'Copy of {sample.name}')
+
+    def test_data_duplication(self):
+        data = self.res.run(slug='test-sleep-progress', input={'t': 1})
+        # Let's not wait for processing to finish and
+        # check the expected exception to be raised.
+        with self.assertRaisesRegex(ResolweServerError, 'done or error status to be duplicated'):
```
JureZmrzlikar: Why do you expect error during duplication?

hadalin: Read the comment above :)

resdk/tests/functional/duplication/e2e_duplication.py
hadalin: This test is there to check what happens if permission error is raised. @dblenkus suggested adding a `PermissionError(ResolweServerError)` to `exceptions.py ` and raising that if resolwe returns permission error.

resdk/tests/functional/duplication/e2e_duplication.py
hadalin: Duplication is allowed only for data objects that have `OK` or `ER` status. In other cases exception is raised and since we only care if endpoint works as expected, we check that.

resdk/tests/functional/duplication/e2e_duplication.py
JureZmrzlikar: OK, this sounds fine.

resdk/tests/functional/duplication/e2e_duplication.py
JureZmrzlikar: OK, makes sense.

JureZmrzlikar: Maybe add an explicit comment that error is expected because data will not have an OK/ER status. I am familiar with Resolwe duplication mechanism, but still haven't made the connection.

resdk/tests/functional/duplication/e2e_duplication.py
```diff
@@ -0,0 +1,39 @@
...
+        sample = self.res.sample.create(name='Test sample')
+        duplicate = sample.duplicate()
+        self.assertEqual(duplicate.name, 'Copy of Test sample')
+
+    def test_sample_duplication_perms(self):
+        collection = self.user_res.collection.create()
+        sample = self.user_res.sample.create(name='Test sample')
+        collection.add_samples(sample)
+
+        collection.permissions.remove_user(self.user_res.auth.username, 'add')
```
dblenkus: Use `USER_USERNAME` (form `resdk.tests.functional.base`) instead of `self.user_res.auth.username`

dblenkus: Sidenote for @JureZmrzlikar: we should add constants with permission values, e.g. `view`, `edit`, `add`,...

dblenkus: @hadalin it would be cleaner approach to create objects with `self.res` (admin user) and then grant permissions to `USER_USERNAME`.

resdk/tests/functional/duplication/e2e_duplication.py
```diff
@@ -0,0 +1,39 @@
...
+        # TODO is it ok that no exception is raised when duplicating a sample
+        # with 'inherit_collection' set to True while the user has no 'add' permission
+        # on the collection?
+        # What happens is that the sample is duplicated just fine, but is also NOT
+        # added to the collection (without an exception being raised).
```
dblenkus: This should be fixed in Resolwe. @hadalin can you do this?

resdk/tests/functional/duplication/e2e_duplication.py
JureZmrzlikar: Regarding constants for view, edit, add, ... I'd rather spend effort polishing this at later time -when permissions changes take place.

dblenkus: >  at later time

That's why it is a sidenote :)

hadalin: >it would be cleaner approach to create objects with self.res

Isn't it required to create a sample like this `sample = self.user_res.sample.create` so that `sample.duplicate()` makes a request as `'user'`?

Also isn't it more obvious what we're trying to test if we _remove_ the permission? If not, then please provide a code snippet, so I can see what you mean.

resdk/tests/functional/duplication/e2e_duplication.py
hadalin: I could, but it's not a priority :)

I'd merge as it is and deal with this later as this is a separate issue.

resdk/tests/functional/duplication/e2e_duplication.py
hadalin: Added an extra comment below.

resdk/tests/functional/duplication/e2e_duplication.py
dblenkus: You can, but it doesn't make sense to add a test that works against expectations :)

hadalin: Well, these _are_ the current expectations :)

Either we

- [ ] keep the test which will fail once the issue is fixed in Resolwe, but we get to keep the test and not forget about it
- [ ] remove the test

resdk/resources/sample.py
```diff
@@ -225,3 +225,22 @@ def is_background(self):
             self._is_background = len(background_relations) > 0  # pylint: disable=len-as-condition
 
         return self._is_background
+
+    def duplicate(self, inherit_collection=False):
+        """Duplicate (make copy of) ``sample`` object.
+
+        :param inherit_collection: If ``True`` then duplicated
+            samples will be added to collections the original sample
```
JureZmrzlikar: Since samples and collection now have m2o instead of m2m relation, this could be simplified. Something like: "If `True` then duplicated samples (and their data) will be added to collection of the original sample."

resdk/resources/collection.py
```diff
@@ -185,3 +185,14 @@ def relations(self):
             self._relations = self.resolwe.relation.filter(collection=self.id)
 
         return self._relations
+
+    def duplicate(self):
+        """Duplicate (make copy of) ``collection`` object.
+
+        :return: Duplicated collection
+        """
+        if self.id is None:
```
JureZmrzlikar: The concept of checking the existence of `id` was so common that we made a [assert_object_exists](https://github.com/genialis/resolwe-bio-py/blob/master/resdk/utils/decorators.py#L24) decorator for it. Please, use it here and in all other `duplicate()` methods.

hadalin:
@JureZmrzlikar I haven't followed what changes were made in Resolwe. E2E tests seems to be failing because of `Bad Request: /api/data/duplicate`. Would you mind giving pointers what needs to be fixed?

PR #196, opened 2019-08-18
Fix ResdkQuery.get

JureZmrzlikar:
This fixes the case where slug is equal (but version is not) for
multiple objects e.g. for Process or DescriptorSchema.

### DIFF
--- a/resdk/query.py
+++ b/resdk/query.py
@@ -13,6 +13,8 @@ import copy
 import logging
 import operator
 
+from resdk.resources import DescriptorSchema, Process
+
 
 class ResolweQuery:
     """Query resource endpoints.
@@ -281,6 +283,12 @@ class ResolweQuery:
 
             arg = args[0]
             kwargs = {'id': arg} if str(arg).isdigit() else {self.slug_field: arg}
+
+        if self.slug_field in kwargs:
+            if issubclass(self.resource, (Process, DescriptorSchema)):
+                kwargs['ordering'] = kwargs.get('ordering', '-version')
+
+            kwargs['limit'] = kwargs.get('limit', 1)
 
         new_query = self._clone()
         new_query._add_filter(kwargs)  # pylint: disable=protected-access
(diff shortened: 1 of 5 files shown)


resdk/query.py
```diff
@@ -280,7 +280,10 @@ def get(self, *args, **kwargs):
                 raise ValueError('Non-keyworded arguments cannot be combined with keyworded ones.')
 
             arg = args[0]
-            kwargs = {'id': arg} if str(arg).isdigit() else {self.slug_field: arg}
+            if str(arg).isdigit():
+                kwargs = {'id': arg}
+            else:
+                kwargs = {self.slug_field: arg, 'ordering': '-version', 'limit': 1}
```
dblenkus: This does not behave the same if `slug` is passed as a keyword or positional argument.

dblenkus:
And you should also update [THIS](https://github.com/genialis/resolwe-bio-py/blob/master/resdk/resolwe.py#L191) accordingly.

resdk/query.py
```diff
@@ -280,7 +280,15 @@ def get(self, *args, **kwargs):
                 raise ValueError('Non-keyworded arguments cannot be combined with keyworded ones.')
 
             arg = args[0]
-            kwargs = {'id': arg} if str(arg).isdigit() else {self.slug_field: arg}
```
dblenkus: Revert this to the original.

resdk/query.py
```diff
@@ -280,7 +280,15 @@ def get(self, *args, **kwargs):
...
+            if str(arg).isdigit():
+                kwargs = {'id': arg}
+            else:
+                kwargs = {self.slug_field: arg}
+        if self.slug_field in kwargs:
```
dblenkus: Add a new line before this if.

resdk/query.py
JureZmrzlikar: Done.

resdk/query.py
JureZmrzlikar: Done.

PR #198, opened 2019-09-17
Fix documentation for Resolwe.run

romunov:
The documentation was not synced how `collection` parameter is treated. Below is a suggested fix.

### DIFF
--- a/resdk/resolwe.py
+++ b/resdk/resolwe.py
@@ -267,7 +267,8 @@ class Resolwe:
         :param dict input: Input values
         :param dict descriptor: Descriptor values
         :param str descriptor_schema: A valid descriptor schema slug
-        :param list collection: Collection into which data object should be included
+        :param int/resource collection: Collection resource or it's id
+            into which data object should be included
         :param str data_name: Default name of data object
 
         :return: data object that was just created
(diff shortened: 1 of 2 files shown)


resdk/resolwe.py
```diff
@@ -267,7 +267,7 @@ def run(self, slug=None, input={}, descriptor=None,  # pylint: disable=redefined
         :param dict input: Input values
         :param dict descriptor: Descriptor values
         :param str descriptor_schema: A valid descriptor schema slug
-        :param list collection: Collection into which data object should be included
+        :param int collection: Collection (id) into which data object should be included
```
JureZmrzlikar: It can be both: `Collection` resource or colection id. This might be longer but also more explicit: "Collection resource or it's id into which data object should be included"

romunov: Good suggestion. I edited to include the resource as well. Not sure if there's a better way of listing several possible values.

resdk/resolwe.py
```diff
@@ -267,7 +267,7 @@ def run(self, slug=None, input={}, descriptor=None,  # pylint: disable=redefined
         :param dict input: Input values
         :param dict descriptor: Descriptor values
         :param str descriptor_schema: A valid descriptor schema slug
-        :param list collection: Collection into which data object should be included
+        :param int/resource collection: Collection resource or it's id into which data object should be included
```
dblenkus: Docstrings should be wrapped at 72 characters.

romunov: Done.

PR #200, opened 2019-11-11
Define ResolweQuery attributes in Resolwe class

JureZmrzlikar:
This fix removes linter warnings of type: "Resolwe class does not
have a `data` member". The warnings are cuased since these
attributes are not defined in __init__, but in
_initialize_queries.

### DIFF
--- a/resdk/resources/collection.py
+++ b/resdk/resources/collection.py
@@ -31,8 +31,6 @@ class BaseCollection(BaseResolweResource):
         'description', 'descriptor', 'descriptor_schema', 'settings', 'tags',
     )
 
-    ALL_PERMISSIONS = ['view', 'download', 'add', 'edit', 'share', 'owner']
-
     def __init__(self, resolwe, **model_data):
         """Initialize attributes."""
         self.logger = logging.getLogger(__name__)
--- a/resdk/resources/permissions.py
+++ b/resdk/resources/permissions.py
@@ -2,6 +2,7 @@
 import copy
 from collections import defaultdict
 
+from ..constants import ALL_PERMISSIONS
 from .utils import is_group, is_user
 
 
@@ -234,7 +235,7 @@ class PermissionsManager:
         perms = copy.copy(perms)
 
         # Order known permissions by importance.
-        for known_perm in ['owner', 'share', 'edit', 'add', 'download', 'view']:
+        for known_perm in ALL_PERMISSIONS:
             if known_perm in perms:
                 perms.remove(known_perm)
                 perms.insert(0, known_perm)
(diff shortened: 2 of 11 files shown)


resdk/resources/collection.py
```diff
@@ -31,7 +31,7 @@ class BaseCollection(BaseResolweResource):
         'description', 'descriptor', 'descriptor_schema', 'settings', 'tags',
     )
 
-    ALL_PERMISSIONS = ['view', 'download', 'add', 'edit', 'share', 'owner']
+    ALL_PERMISSIONS = ['view', 'edit', 'share', 'owner']
```
mstajdohar: Could we define those on `BaseResolweResource` or make it constant?

resdk/resources/permissions.py
```diff
@@ -234,7 +234,7 @@ def _perms_to_string(self, perms):
         perms = copy.copy(perms)
 
         # Order known permissions by importance.
-        for known_perm in ['owner', 'share', 'edit', 'add', 'download', 'view']:
+        for known_perm in ['owner', 'share', 'edit', 'view']:
```
mstajdohar: We could abstract this bit if we made permissions a constant.

resdk/tests/functional/docs/e2e_docs.py
```diff
@@ -70,9 +70,9 @@ def upload_reads(self, res):
             slug='upload-fastq-single',
             input={'src': os.path.join(TEST_FILES_DIR, 'reads.fastq.gz')},
         )
-        self.set_slug_and_make_public(reads, self.reads_slug, permissions=['view', 'download'])
+        self.set_slug_and_make_public(reads, self.reads_slug, permissions=['view'])
         self.set_slug_and_make_public(
-            reads.sample, self.sample_slug, permissions=['view', 'download']
+            reads.sample, self.sample_slug, permissions=['view']
```
mstajdohar: Can we make this one line now?

resdk/resources/collection.py
JureZmrzlikar: Defined in `BaseResolweResource`.

resdk/resources/permissions.py
JureZmrzlikar: What do you mean?
`for known_perm in self.all_permissions:` (and self.all_permissions are ordered by importance)

mstajdohar: For examples, define `ALL_PERMISSIONS` in `constants.py`.

JureZmrzlikar: Aha, OK. Done.

PR #201, opened 2019-11-14
Add methods to list users with certain permission

JureZmrzlikar:


resdk/resources/base.py
```diff
@@ -235,6 +235,29 @@ def permissions(self):
...
+        holders = []
+        self.permissions.fetch()
+
+        for item in self.permissions._permissions:  # pylint: disable=protected-access
+            if perm in item['permissions']:
+                holders.append(item.get('name', 'Public'))
+
+        return holders
+
+    def list_owners(self):
```
dblenkus: This doesn't list users, but returns them, so more appropriate name would be `get_owners`. Considering the implementation of other similar cases in the ReSDK it may also make sense to make this a property.

docs/CHANGELOG.rst
```diff
@@ -19,6 +19,8 @@ Changed
 Added
 -----
 - Add duplicate method to collection, sample and data resources
+- Add methods to list users with certain permission: ``list_owners()``,
```
dblenkus: Where have you added these methods?

resdk/resources/base.py
```diff
@@ -235,6 +235,29 @@ def permissions(self):
...
         return self._permissions
 
+    def _list_holders_with_perm(self, perm):
+        """Get users/group/public names that have ``perm`` perm."""
+        holders = []
+        self.permissions.fetch()
+
+        for item in self.permissions._permissions:  # pylint: disable=protected-access
+            if perm in item['permissions']:
+                holders.append(item.get('name', 'Public'))
```
dblenkus: It would be nice (and consistent) if you return `User` objects instead of just names.

resdk/resources/base.py
dblenkus: Have you thought about what makes more sense, adding this method to the `<object>` or to the `<object>.permissions`?

resdk/resources/base.py
```diff
@@ -235,6 +235,29 @@ def permissions(self):
...
+
+        return holders
+
+    def list_owners(self):
+        """List owners."""
```
dblenkus: Make the docstring more informative, e.g. `Return list of users with ``owner`` permission.`

resdk/resources/base.py
JureZmrzlikar: My initial implementation was in `<object>.permissions`, yes. I changed my mind since it made more sense to say `Data.get_owners()` than `Data.permissions.get_owners()`. But the benefit of the `<object>.permissions` is that it clearly separates the "permission-handling" to `PermissionManager` class. What do you think?

If we decide to keep it in `<object>` i think making this a propery (`owners`, `editors`, ...) is a good choice. If we go with `<object>.permissions` i think it is more understandable if we use `get_owners()`.

docs/CHANGELOG.rst
JureZmrzlikar: Will fix when we decide :-)

resdk/resources/base.py
JureZmrzlikar: Good point. But returning the "full" user would require another request for each user... We can create user instances "partially", like in `BaseResolweReource.contributor`. But this can be tricky since User instance will be missing some of the attributes that it should have.

JureZmrzlikar:
@dblenkus This is ready for review. The methods now return User/Group instances. They also include username which is now available from Resolwe API ([Resolwe PR](https://github.com/genialis/resolwe/pull/654))

resdk/resources/permissions.py
```diff
@@ -269,3 +276,43 @@ def __repr__(self):
...
+    def _get_holders_with_perm(self, perm):
+        """Get Users/Group/public names that have ``perm`` perm."""
+        # Prevent circular imports
+        from .user import Group, User
+
+        self.fetch()
+
+        holders = []
+        for item in self._permissions:  # pylint: disable=protected-access
+            if perm in item['permissions']:
```
dblenkus: You can use [filter](https://book.pythontips.com/en/latest/map_filter.html#filter) instead of this if.

resdk/resources/permissions.py
```diff
@@ -269,3 +276,43 @@ def __repr__(self):
...
+                        User(
+                            self.resolwe,
+                            id=item['id'],
+                            first_name=item['name'],
+                            username=item['username'],
+                        ),
+                    )
+                elif item['type'] == 'group':
+                    holders.append(Group(self.resolwe, id=item['id'], name=item['name']))
+                else:
```
dblenkus: Change to `elif item['type'] == 'public'` for a sanity check.

resdk/tests/functional/permissions/e2e_permissions.py
```diff
@@ -6,50 +6,54 @@
...
-        collection_admin.permissions.add_user(USER_USERNAME, 'view')
+        self.test_collection.permissions.add_user(USER_USERNAME, 'view')
 
         # User can see the collection, but cannot edit it.
-        collection_user = self.user_res.collection.get(collection_admin.id)
-        collection_user.name = 'New test collection'
+        user_collection = self.user_res.collection.get(self.test_collection.id)
         with self.assertRaises(ResolweServerError):
-            collection_user.save()
+            user_collection.name = 'Different name'
```
dblenkus: This line doesn't need to be inside the context manager.

resdk/tests/functional/permissions/e2e_permissions.py
```diff
@@ -6,50 +6,54 @@
...
 
-        collection_admin.permissions.remove_user(USER_USERNAME, 'edit')
+        self.test_collection.permissions.remove_user(USER_USERNAME, 'edit')
 
         # Edit permission is removed again.
-        collection_user = self.user_res.collection.get(collection_admin.id)
-        collection_user.name = 'Another collection'
         with self.assertRaises(ResolweServerError):
-            collection_user.save()
+            user_collection.name = 'Different name 2'
```
dblenkus: Move this line out of the context manager.

resdk/tests/functional/permissions/e2e_permissions.py
```diff
@@ -6,50 +6,54 @@
...
+        self.test_collection.permissions.add_user(USER_USERNAME, ['edit', 'view'])
+        self.test_collection.permissions.add_public('view')
 
-        self.assertEqual(len(collection.permissions.editors), 2)
-        self.assertEqual(collection.permissions.editors[0].get_name(), "admin")
-        self.assertEqual(collection.permissions.editors[1].get_name(), "user")
+        self.assertEqual(len(self.test_collection.permissions.owners), 1)
+        self.assertEqual(self.test_collection.permissions.owners[0].get_name(), "admin")
 
+        self.assertEqual(len(self.test_collection.permissions.editors), 2)
```
dblenkus: These lines are a part of the previous commit.

resdk/resources/permissions.py
```diff
@@ -269,3 +276,43 @@ def __repr__(self):
...
+                    holders.append(User(self.resolwe, username='public', first_name="Public"))
+        return holders
+
+    @cachedproperty
```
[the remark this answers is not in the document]
JureZmrzlikar: Why not? This is not a random third party library. It is 10 times more popular than for example `slumber` package that we heavily use. And why do you think it complicates the code? I see it as a very nice simplification. The aim here was to try it and see if it works OK. Then we can apply this pattern to other parts of the code.

resdk/resources/permissions.py
JureZmrzlikar: I believe this approach is clearer than using filter, but this might be a matter of taste.

resdk/resources/permissions.py
JureZmrzlikar: :+1:

resdk/tests/functional/permissions/e2e_permissions.py
JureZmrzlikar: They were created in previous commit, but were edited as part of this comit.

resdk/resources/permissions.py
```diff
@@ -71,6 +73,11 @@ def _set_permissions(self, action, perms, who_type, who=None):
     def clear_cache(self):
         """Clear cache."""
         self._permissions = None
+        # ``boltons.cacheutils.cachedproperty`` is saved/cached in ``__dict__``
+        # after first call, it needs to be cleared there too.
+        for attribute in ["owners", "editors", "viewers"]:
+            if attribute in self.__dict__:
+                del self.__dict__["owners"]
```
dblenkus: `"owners"` -> `attribute`

JureZmrzlikar: Oh, nice catch, thanks!

resdk/resources/permissions.py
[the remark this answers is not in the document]
dblenkus: Each such addition to the codebase increase barrier to entry and multiplies the chance that something gets wrong. It is definitely useful to check such implementations and reuse bits of them, but even this must be done with care as it can get complicated very quickly (e.g. modifying the `__dict__`).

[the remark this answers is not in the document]
JureZmrzlikar: Thanks for this feedback. You've convinced me :-) I've rewritten the code to mimic the current style of caching.

PR #202, opened 2019-11-14
Add iterate method to ResolweQuery

JureZmrzlikar:


resdk/query.py
```diff
@@ -352,3 +352,13 @@ def search(self, text):
         new_query._add_filter({self.resource.full_search_paramater: text})
         # pylint: enable=protected-access
         return new_query
+
+    def chunked_all(self, chunk_size=100):
```
dblenkus: I would name this method iterator.

resdk/query.py
```diff
@@ -352,3 +352,13 @@ def search(self, text):
...
+        """Iterate through all objects."""
+        max_id = self.filter(ordering='-created', limit=1)[0].id
+        ids = list(range(1, max_id +  1))
+        sublists = [ids[i:i + chunk_size] for i in range(0, len(ids), chunk_size)]
```
dblenkus: You should use pagination instead of sublists. That way this will also work on queries with already applied filters.

To simplify the implementation, you can order objects by `id` in the first implementation. But you should raise an error if ordering is already applied to the query (btw. it would be nice to add `order_by` method at some point).

To speed up the retrieval of later pages (and to do a workaround for Elasticsearch limitations), you should filter each query by the id of the last element of previous query (i.e. `.filter(id__gt=last.id)`). But you should check how filters and queries work if two filters by `id` are applied and handle that situation appropriately.

resdk/query.py
```diff
@@ -352,3 +352,29 @@ def search(self, text):
...
+            raise ValueError(
+                "Parameter 'limit' should not be used in combination with method iterate."
+            )
+        if self._offset is not None:
+            raise ValueError(
+                "Parameter 'offset' should not be used in combination with method iterate."
+            )
+
+        count = self.count()
+        self._offset = 0
```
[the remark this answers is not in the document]
JureZmrzlikar: Good point, thanks.

PR #205, opened 2020-04-02
Update filtering docs

JureZmrzlikar:


docs/files/tutorial-get.py
```diff
@@ -20,8 +20,8 @@
 # Filter by using several fields:
 res.data.filter(
   status='OK',
-  created__gt='2018-10-01',
-  created__lt='2025-11-01',
+  created__gt='2018-10-01 01:23:45',
```
dblenkus, line 23: Why don't you use `datetime.datetime(2018, 10, 1)`? It looks more natural and easier to remember.

JureZmrzlikar: Done.

docs/tutorial-get.rst
```diff
@@ -63,7 +63,7 @@ But the real power of the ``filter()`` method is in combining multiple
 parameters:
 
 .. literalinclude:: files/tutorial-get.py
-  :lines: 20-27
+  :lines: 20-29
```
dblenkus, line 66: Does this also break all other references further down the file?

JureZmrzlikar: Yes :-)

PR #210, opened 2020-05-07
Update linting tools

JureZmrzlikar:


tox.ini
```diff
@@ -33,13 +33,13 @@ commands =
     docs: python setup.py build_sphinx --fresh-env --warning-is-error
 
     # Linter commands:
-    linters: pylint resdk
+    linters: pylint src
```
dblenkus: You also want to lint tests directory.

setup.py
```diff
@@ -28,9 +28,8 @@
     url=about['__url__'],
     license=about['__license__'],
     # Exclude tests from built/installed package.
-    packages=setuptools.find_packages(
-        exclude=['tests', 'tests.*', '*.tests', '*.tests.*']
-    ),
+    packages=setuptools.find_packages("src", exclude=["tests", "tests.*", "*.tests", "*.tests.*"]),
```
dblenkus, line 31: Is the exclude required as you moved tests out of `src`?

JureZmrzlikar: I'm not sure, i've copied this from [DRF-reactive](https://github.com/genialis/django-rest-framework-reactive/blob/master/setup.py) that has the same structure :-)

JureZmrzlikar: Thanks for this comment. If i remove it all works fine, so i'll just leave it without exclude.

PR #213, opened 2020-05-21
Add cookies to redirects

gregorjerse:
Adding cookies in custom authenticator does not add cookies to the
automatically generated queries when redirect is followed.

This fix uses session to store auth cookies and only sets headers on custom
authenticator call.

### DIFF
--- a/src/resdk/resolwe.py
+++ b/src/resdk/resolwe.py
@@ -98,8 +98,11 @@ class Resolwe:
     feature = None
     mapping = None
 
+    session = None
+
     def __init__(self, username=None, password=None, url=None):
         """Initialize attributes."""
+        self.session = requests.Session()
         if url is None:
             # Try to get URL from environmental variable, otherwise fallback to default.
             url = os.environ.get("RESOLWE_HOST_URL", DEFAULT_URL)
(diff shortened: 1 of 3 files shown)


src/resdk/resolwe.py
```diff
@@ -97,9 +97,11 @@ class Resolwe:
     group = None
     feature = None
     mapping = None
+    session = None
```
JureZmrzlikar, line 100: Can you add an empty line above? Now, it appears as session is one of the resources...

src/resdk/resolwe.py
```diff
@@ -485,6 +493,7 @@ class ResAuth(requests.auth.AuthBase):
     def __init__(self, username=None, password=None, url=DEFAULT_URL):
         """Authenticate user on Resolwe server."""
         self.logger = logging.getLogger(__name__)
+        self.cookies = {}
```
JureZmrzlikar, line 496: Why define empty dict here if you re-create it in the lines below (line 524)?

src/resdk/resolwe.py
gregorjerse: Sure. Actually I added session here just to make tests pass.
I can also remove it and add it in test methods.

src/resdk/resolwe.py
gregorjerse: To always have cookies defined. If no username and password is given, code will return None before it reaches the last line in `__init__`.

PR #215, opened 2020-05-25
Improve docstring of ResolweQuery.iterat method

JureZmrzlikar:


src/resdk/query.py
```diff
@@ -364,8 +364,13 @@ def iterate(self, chunk_size=100):
 
         This can come handy when one wishes to iterate through hundreds or
         thousands of objects and would otherwise get "504 Gateway-timeout".
+
+        However there are certain limitations. This method cannot be
+        used if ResolweQuery has set ``limit``, ``offset`` or
```
mstajdohar, line 369: I suggest: The method cannot be used together with the following filters: ``limit``, ``offset`` and ``ordering``, and will raise a ``ValueError``.

PR #217, opened 2020-09-14
Add permissions.copy_from() method

JureZmrzlikar:


docs/CHANGELOG.rst
```diff
@@ -5,6 +5,15 @@ Change Log
 All notable changes to this project are documented in this file.
 
 
+==========
+Unreleased
+==========
+
+Added
+-----
+- Add ``permissions.copy_from()`` method
```
dblenkus, line 14: I would name it `<resource>.permissions.copy_from(<source>)` and add a short description of what the method does.

dblenkus (approved):

docs/CHANGELOG.rst
JureZmrzlikar: Good suggestion, but I've modified it a bit.

PR #220, opened 2020-09-24
Support login with email (MVP)

JureZmrzlikar:


src/resdk/resolwe.py
```diff
@@ -502,7 +503,10 @@ def __init__(self, username=None, password=None, url=DEFAULT_URL):
         if not username and not password:
             return
 
-        payload = {"username": username, "password": password}
+        if is_email(username):
```
mstajdohar, line 506: How about something like this?

```
key = "email" if is_email(username) else "username"
payload = {key: username, "password": password}
```

mstajdohar (approved):
Splendid, thank you

mstajdohar:
Do we need to update the docs too?

src/resdk/resolwe.py
JureZmrzlikar: Good suggestion, thanks!

JureZmrzlikar:
> Do we need to update the docs too?
I've updated the login prompt.

PR #221, opened 2020-09-25
Fix broken sample assignment

JureZmrzlikar:
If updating Data.sample property, one needs to post updates via
entity key, not sample.

### DIFF
--- a/src/resdk/resources/base.py
+++ b/src/resdk/resources/base.py
@@ -132,6 +132,8 @@ class BaseResource:
                     payload[field_name] = self._dehydrate_resources(
                         getattr(self, field_name)
                     )
+            if "sample" in payload:
+                payload["entity"] = payload.pop("sample")
 
             if payload:
                 response = self.api(self.id).patch(payload)
@@ -146,6 +148,9 @@ class BaseResource:
                 for field_name in field_names
                 if getattr(self, field_name) is not None
             }
+
+            if "sample" in payload:
+                payload["entity"] = payload.pop("sample")
 
             response = self.api.post(payload)
             self._update_fields(response)
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -13,6 +13,10 @@ Added
 -----
 - Support login with email
 
+Fixed
+-----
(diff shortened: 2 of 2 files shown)


src/resdk/resources/base.py
```diff
@@ -132,6 +132,9 @@ def assert_fields_unchanged(field_names):
                     payload[field_name] = self._dehydrate_resources(
                         getattr(self, field_name)
                     )
+            if "sample" in payload:
+                payload["entity"] = copy.deepcopy(payload["sample"])
```
dblenkus, line 136: Why do you need a deep copy? The original reference (i.e., `payload["sample"]`) is deleted anyway so you can use the same object. 

You can use `payload.pop("sample")` instead of `payload["sample"]` + `del payload["sample"]`

mzganec (approved):
I tested the code. The assignment works with both `data.sample = sample` and `data.sample = sample.id`. Thanks!

src/resdk/resources/base.py
dblenkus: Is the same thing also needed in the `else` below? It would be nicer to move this into the `Data.save` method and handle it before this method is called.

JureZmrzlikar: > Is the same thing also needed in the else below?
Yes it is, thanks!

> It would be nicer to move this into the Data.save method and handle it before this method is called.
True, but to achieve this, one would need to mess with `self._original_values`, which is kinda against what the "original values" should be. The "proper" way is to refactor save method to enable better subclassing but i hoped i could get away without that :-D

dblenkus: I see it would be much more challenging then I thought. 

Where does the `sample` name come from? I think that the right solution would be to use `entity` everywhere in the codebase and add `sample` property to the `Data` class as an alisas for `entity`. Would that be feasible?

JureZmrzlikar:
I'll merge this to fix the issue raised by. I'm leaving the task of:
> I think that the right solution would be to use entity everywhere in the codebase and add sample property to the Data class as an alisas for entity.

for some later time.

PR #223, opened 2020-10-28
Fix broken get requests

gregorjerse:
Use session.get instead of requests.get otherwise the authentication
does not work and requests fail with HTTP status "403 Forbidden".


JureZmrzlikar (approved):
Thank @gregorjerse , this works for me. Could you also add a changelog entry?

dblenkus (approved):

gregorjerse:
> Thank @gregorjerse , this works for me. Could you also add a changelog entry?

Done.

PR #227, opened 2020-11-17
Add Knowledge-base docs

JureZmrzlikar:


JureZmrzlikar:
I ask:
- @acopar To review the PR technically
- @MiaZganjar and @mzganec to review PR contextually. I would benefit from a second eye and possible improvement suggestions. Thanks.

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
+.. _knowleedge-base:
```
dblenkus, line 1: Would it make sense to move the file to the `topical` subdirectory to mimic the organization of the content?

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+        target_db="UniProtKB",
+        source_species="Homo sapiens",
+        target_species="Homo sapiens",
+    )
+    uniprot_id = mapping.target_id
+
+    # Find mouse ortholog for Human BRCA2 gene:
+    mapping = res.mapping.get(
+        relation_type="ortholog",
+        source_id="ENSG00000139618",
```
dblenkus, line 101: I know I am not a biologist, but `BRCA2` and `ENSG00000139618` don't look to similar.

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+    # Feature type, e.g. gene, transcript, exon, ...
+    feature1.type
+
+    # Feature name
+    feature1.name
+
+    # List of feature aliases
+    feature1.aliases
+
+But the real power is in the search / filter capabilities. Let's just
```
dblenkus, line 50: Examples below are only for filtering, so you can omit search.

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+exon, etc. You can search and filter ``Feature`` objects by ``feature``
+endpoint, similarly like ``Data``, ``Sample`` or any other ReSDK resource::
+
+    feature1 = res.feature.get(feature_id="BRCA2")
+
+To examine all attributes of a ``Feature``, see the :ref:`reference`.
+Here we will list a few::
+
+    # Get the feature:
+    feature1 = res.feature.get(feature_id="BRCA2")
```
dblenkus, line 30: Change `feature1` to `feature`.

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+        type="gene",
+        species="Homo sapiens",
+    )
+    gene_symbols = [g.name for g in genes]
+
+
+Mapping
+=======
+Mapping is a *connection* between two features. Features can be related
+in various ways - we have different types of relations between features.
```
dblenkus, line 73: Is "we have different types of relations between features" really necessary? I think it only rephrases what the first part of the sentence already tells you.
I would remove it and change the start of the next sentence to "The type of a mapping is indicated..."

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+This type is indicated by ``relation_type`` attribute. It can be one of
+the following options:
+
+    - Type: ``crossdb``. Two features from different sources (databases)
```
dblenkus, line 77: I would omit `Type: ` and put a dot at the end of each sentence.

docs/knowledge-base.rst
dblenkus: There is also no need for "But" at the beginning of the sentence.

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+    - finding all genes of type ``protein_coding``
+    - find all transcripts of gene ``FHIT``
+    - converting ``gene_id`` to ``gene_symbol``
+    - ...
+
+
+Feature
+=======
+``Feature`` object represents a genomic feature: a gene, a transcript, an
+exon, etc. You can search and filter ``Feature`` objects by ``feature``
```
dblenkus, line 21: Maybe "query" instead of "search and filter".

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+        feature_id__in=gene_ids,
+        type="gene",
+        species="Homo sapiens",
+    )
+    gene_symbols = [g.name for g in genes]
+
+
+Mapping
+=======
+Mapping is a *connection* between two features. Features can be related
```
dblenkus, line 72: "Features can be..." -> "They can be..."

dblenkus (commented):
Looks good in general, but I don't like the tone of the writing. E.g. just list some, list a few, only list some,...It would be much more convincing if you write something like "You can find a complete list in the reference, so let's just list commonly used use-cases."

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+        source="ENSEMBL",
+    ).count()
+
+    # Convert all gene_id's in a list ``gene_ids`` to gene symbols::
+    genes = res.feature.filter(
+        feature_id__in=gene_ids,
+        type="gene",
+        species="Homo sapiens",
+    )
+    gene_symbols = [g.name for g in genes]
```
mzganec, line 67: This is a good example of how not to do a conversion! I would actually keep this example in the docs with a big warning, so that the readers avoid this mistake. Because it's certainly tempting to do it like this. The problem, of course, is that there is no guarantee that `gene_symbols` will appear in the same order as `gene_ids`. I presume that the order is going to be the same as in the database, but @dblenkus probably knows much more about this. So I welcome that we show this example but of course also show how to do the conversion correctly.

docs/knowledge-base.rst
mzganec: I prefer @JureZmrzlikar's phrasing here.

mzganec (approved):
Please fix the misguiding example. Other than that the docs look good. I have learned something new here and intend to use this functionality in the future.

docs/knowledge-base.rst
```diff
@@ -0,0 +1,107 @@
...
+
+    # Count number of Human "protein-conding" transcripts in ENSEMBL database
+    res.feature.filter(
+        species="Homo sapiens",
+        type="transcript",
+        subtype="protein-coding",
+        source="ENSEMBL",
+    ).count()
+
+    # Convert all gene_id's in a list ``gene_ids`` to gene symbols::
```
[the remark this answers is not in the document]
mzganec: Actually, we want here _gene IDs_ instead of _gene_id's_. So lose the underscore as well.

docs/knowledge-base.rst
JureZmrzlikar: Maybe, then it would also make sense to put tutorial source files into their own folder. But then they would mix up with `files` and `images` folders that have nothing to do with content hierarchy. 

Currently I prefer flat structure, but as soon as number of files goes above ~20 we will need some structure.

docs/knowledge-base.rst
JureZmrzlikar: Good point, thanks. I've fixed the example.

JureZmrzlikar:
Thanks @dblenkus  and @mzganec for your fast response. I've modified the writing with your suggestions. Please review again.

dblenkus (commented):
Looks good to me!

docs/knowledge-base.rst
```diff
@@ -0,0 +1,117 @@
...
+        source="ENSEMBL",
+    ).count()
+
+    # Convert all gene IDs in a list ``gene_ids`` to gene symbols::
+    genes = res.feature.filter(
+        feature_id__in=gene_ids,
+        type="gene",
+        species="Homo sapiens",
+    )
+    mapping = {g.feature_id: g.name for f in genes}
```
mzganec, line 66: `for g in genes` 😉

PR #232, opened 2021-01-15
Add CollectionTables docs

JureZmrzlikar:


docs/collection-tables.rst
```diff
@@ -0,0 +1,89 @@
...
+rows represents samples and columns represent genes. Class
+``CollectionTables`` gives you just that (and more).
+
+We will present the functionality of ``CollectionTables`` through an
+example. We will:
+
+- Create an instance of ``CollectionTables`` and examine it's attributes
+- Fetch raw expression counts and only keep `TIS genes`_ with sufficient
+  expression
+- Compute ``log2(TPM + 1)`` of expression values and visualize samples
```
mstajdohar, line 18: Normalize expression values (log-transform) and...

docs/collection-tables.rst
```diff
@@ -0,0 +1,89 @@
+.. _collection-tables:
+
+=================
+Collection Tables
```
mstajdohar, line 4: Uninformative title. I would not find this functionality if I didn't know where to look. I can't think of a better alternative either.

docs/collection-tables.rst
```diff
@@ -0,0 +1,89 @@
...
+Imagine you are modelling gene expression data from a given collection.
+Ideally, you would want all expression values organized in a table where
+rows represents samples and columns represent genes. Class
+``CollectionTables`` gives you just that (and more).
+
+We will present the functionality of ``CollectionTables`` through an
+example. We will:
+
+- Create an instance of ``CollectionTables`` and examine it's attributes
+- Fetch raw expression counts and only keep `TIS genes`_ with sufficient
```
mstajdohar, line 16:
```
Fetch raw expressions and select `TIS signature genes`_ with sufficient coverage
```

docs/collection-tables.rst
```diff
@@ -0,0 +1,89 @@
...
+- Compute ``log2(TPM + 1)`` of expression values and visualize samples
+  in a simple PCA plot
+
+.. _`TIS genes`: https://translational-medicine.biomedcentral.com/articles/10.1186/s12967-019-2100-3
+
+First, connect to a Resolwe server, pick a collection and create
+and instance of ``CollectionTables``::
+
+    import resdk
+    res = resdk.Resolwe(url='https://app.genialis.com/')
```
mstajdohar, line 27: Note: we should have `app.genialis.com` by default.

docs/collection-tables.rst
```diff
@@ -0,0 +1,89 @@
...
+    res.login()
+    collection = res.collection.get("sum149-fresh-for-rename")
+    ct = resdk.CollectionTables(collection)
```
mstajdohar, line 30: Call it `sum149` instead of `ct`. This improves code readability, especially when you work with more than one collection.

docs/collection-tables.rst
```diff
@@ -0,0 +1,89 @@
...
+
+
+.. note::
+
+  Expressions and metadata are cached in memory as well as on disk. At
+  each time they are re-requested a check is made that local and server side
+  of data is synced. If so, cached data is used. Otherwise, new data
+  will be pulled from server.
+
+In our example we will only work with a set of `TIS genes`_::
```
mstajdohar, line 59:
```
`TIS signature genes`_
```

docs/collection-tables.rst
```diff
@@ -0,0 +1,89 @@
...
+
+    TIS_GENES = ["CD3D", "IDO1", "CIITA", "CD3E", "CCL5", "GZMK", "CD2", "HLA-DRA", "CXCL13", "IL2RG", "NKG7", "HLA-E", "CXCR6", "LAG3", "TAGAP", "CXCL10", "STAT1", "GZMB"]
+
+We will identify low expressed genes and only keep the ones with average raw
+expression above 1::
```
mstajdohar, line 64: Above 20 make more sense, but I assume you don't get enough genes with sufficient coverage then.

mstajdohar (approved):
Looks good. Some minor comments.

docs/collection-tables.rst
JureZmrzlikar: I am against (an)other title, since inventing another term just for the purpuse of title will only increase confusion. If we do decide for a more meaningful rename, we should also rename `CollectionTable` class name.

docs/collection-tables.rst
JureZmrzlikar: Current default is `localhost`. Are we sure about app being the default url?

docs/collection-tables.rst
JureZmrzlikar: Ok.

docs/collection-tables.rst
JureZmrzlikar: Better bad PCA plot than introducing bad filtering practice. Fixed to 20.

mstajdohar: How many genes are you left with?

docs/collection-tables.rst
mstajdohar: @dblenkus, what's your take on that?

docs/collection-tables.rst
JureZmrzlikar: Five (5)

JureZmrzlikar: If i set threshold to 1 then 11 genes remain.

mstajdohar: I assume that's because we have a low number of samples... Five is okay.

PR #233, opened 2021-01-22
Speedup of CollectionTables download with async

JureZmrzlikar:


JureZmrzlikar:
Some basic metrics to get the feel for speedup. We did the test on a collection with 173 samples:
For admins: before 97s, now 22s, cca. ~4 fold increase in speed 
For normal users: before 230s, now 150s, cca. ~1.5 fold increase in speed 

The difference between admins and normal users comes from permisson checks that are very slow. In the future we will speedup the backend even more (especially the permissions checks) and we expect significant improvement also for normal users. PR is already made: [resolwe#781](https://github.com/genialis/resolwe/pull/813)

src/resdk/collection_tables.py
```diff
@@ -417,27 +433,51 @@ def _expression_file_url(self, data: Data, exp_type: str) -> str:
...
-            )
-            response.raise_for_status()
-            with BytesIO() as f:
-                f.write(response.content)
-                f.seek(0)
-                df_ = pd.read_csv(f, sep="\t", compression="gzip")
-                df_ = df_.set_index("Gene").T
-                df_.index = [data._original_values["entity"]["name"]]
-                df_list.append(df_)
+        for i in tqdm(range(0, len(self._data), EXP_ASYNC_CHUNK_SIZE)):
```
robertcv, line 463: You should add back `tqdm` settings so that the user knows what is being downloaded and so that this progress bar is consistent with the gene mapping progress bar.

robertcv (approved):
Besides the one comment, looks good.

src/resdk/collection_tables.py
JureZmrzlikar: Good point, fixed.

src/resdk/collection_tables.py
```diff
@@ -45,6 +48,8 @@
 ]
 
 CHUNK_SIZE = 1000
+# When downloading expressions for large collections, download in chunks
```
dblenkus, line 51: This comment does not explain the constant below.

src/resdk/collection_tables.py
```diff
@@ -263,7 +268,18 @@ def _load_fetch(self, data_type: str) -> pd.DataFrame:
             if data_type == META:
                 data = self._download_metadata()
             else:
-                data = self._download_expressions(data_type)
+                try:
+                    loop = asyncio.new_event_loop()
+                    asyncio.set_event_loop(loop)
```
dblenkus, line 273: You should almost never replace an existing event loop use [`get_event_loop`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop) instead.

src/resdk/collection_tables.py
```diff
@@ -263,7 +268,18 @@ def _load_fetch(self, data_type: str) -> pd.DataFrame:
...
+                    data = loop.run_until_complete(
+                        self._download_expressions(data_type)
+                    )
+                finally:
+                    loop.close()
+                # XXX: When dropping support for Python 3.6 the above
```
dblenkus, line 279: Comments are usually above the code block that they refer to.

src/resdk/collection_tables.py
```diff
@@ -417,27 +433,55 @@ def _expression_file_url(self, data: Data, exp_type: str) -> str:
                 f"Data {data.slug} has multiple expressions of type {exp_type}!"
             )
 
-        return urljoin(self.resolwe.url, f"data/{data.id}/{exp_files[0]}")
+        return f"{data.id}/{exp_files[0]}"
+
+    def _get_exp_urls(self, uris):
```
dblenkus, line 438:
```py
from urllib.parse import urlencode
urlencode({'uri': uris}, doseq=True)
```

dblenkus (approved):

PR #236, opened 2021-03-25
Enable custom callable for reporting download progress

JureZmrzlikar:


dblenkus (approved):

robertcv:
This seems to be a bit overengineered. While I admire your regex expression I also think it could and should be avoided and the simplest solution should be applied. See section `Custom callback` in [docs](https://github.com/tqdm/tqdm#hooks-and-callbacks) where they show a simple example of how to add an external callback.

There is also the question of how to handle the second progress bar (mapping of genes)? Do we use the same callback, do we use a different one, or maybe we just don't add a callback to it because it is just not used that often.

And finally how about when the data is loaded from cache? Shouldn't we also call the callback? While we would probably immediately call it with 100%, at least the behavior would be more consistent.

@JakaKokosar could you provide any insights into how a callback could be used so that this PR can be as helpful as possible.

JureZmrzlikar:
@JakaKokosar Ping.

JakaKokosar:
I am not entirely sure how is this supposed to work. Am I missing something?

```python
In [21]: def callMe():
    ...:     print('Maybe')
    ...:

In [22]: collection = res.collection.get("sum149-fresh-for-rename")

In [23]: sum149 = resdk.CollectionTables(collection, progress_callable=callMe)

In [24]: sum149.rc
Downloading expressions: 100%|████████████████████████████████████████| 1/1 [00:02<00:00,  2.13s/it]
Out[24]:
Ensembl              ENSG00000000003  ENSG00000000005  ENSG00000000419  ENSG00000000457  ENSG00000000460  ENSG00000000938  ...  ERCC-00163  ERCC-00164  ERCC-00165  ERCC-00168  ERCC-00170  ERCC-00171
sample_name                                                                                                                ...
SUM149R_DMSO_12H_R1             1580                0             3946              371             1203                1  ...           0           0           0           0           0           0
SUM149R_DMSO_12H_R2             1391                0             3528              357             1199                0  ...           0           0           0           0           0           0
SUM149R_JQ1_12H_R1              1005                0             4256              443              936                1  ...           0           0           0           0           0           0
SUM149R_JQ1_12H_R2               792                0             3629              373              818                1  ...           0           0           1           0           0           0
SUM149_DMSO_12H_R1              1807                0             3284              468             1573                0  ...           0           0           1           0           0           0
SUM149_DMSO_12H_R2              1321                0             2418              317             1205                1  ...           0           0           0           0           0           0
SUM149_JQ1_12H_R1                923                0             3129              474              546                2  ...           0           0           1           0           0           0
SUM149_JQ1_12H_R2                931                0             3106              476              537                2  ...           0           0           1           0           0           0

[8 rows x 58487 columns]

In [25]: sum149.progress_callable()
Maybe
```

JureZmrzlikar:
@JakaKokosar thanks for responding.

AFAIK, the aim of this callable is reporting progress in some Orange widget, right? Based on my conversation with Marko Toplak, I understood that the "standard" callable in Orange is a function that is called with one argument - number in [0, 1].

So, this should work:

```
def call_me(x):
    """Value of x should be in [0,1]."""
    print(f"\n{x}", flush=True)

collection = res.collection.get("sum149-fresh-for-rename")
sum149 = resdk.CollectionTables(collection, progress_callable=call_me)

# To avoind using cache (and thus forcing the download) also clear cache and restart the instance:
sum149.clear_cache()
sum149 = resdk.CollectionTables(collection, progress_callable=call_me)
sum149.rc
```
And the output in this case is (note the `1.0` line in between):
```
Downloading expressions: 100%|████████████████████████████████████████| 1/1 [00:02<00:00,  2.14s/it]
1.0
Downloading expressions: 100%|████████████████████████████████████████| 1/1 [00:02<00:00,  2.14s/it]
```
Of course, the `call_me` function should do something smarter, e.g. updating progress bar.

p.s. Don't ask why you only get one progress value (1.0) instead of a whole range of them... This is a long story. For bigger collections (> 50 samples) the number of "progress values" increases.

JakaKokosar:
Oh, this makes sense now. LGTM!


This is more in the category "nice to have" but is there a way to disable progress printed on stdout when callable is provided?

mstajdohar (approved):
LGTM. Similar to Jaka, I don't like that the progress is reported to stderr/out when given a callable.

JureZmrzlikar:
OK, thanks for feedback. I've redirected the output to `/dev/null` in case callable is given. Otherwise, all stays the same. Can you check if it's working properly for you @JakaKokosar ?

If all good, i can merge this.

JakaKokosar:
@JureZmrzlikar It works for me, thanks!

PR #247, opened 2021-06-02
Geneset resource

AGregorc:
This is currently still a work in progress for implementing Geneset resource.

What is implemented:
- Geneset constructor, eg.:
    -  Geneset(resolwe, genes=["FHIT", "MYC"], source="UCSC", species="Homo sapiens")
- set-like operations on Geneset objects
- add geneset to the list of Resolwe attributes
    - initial work on res.geneset.get and res.geneset.filter methods
    - create does not work yet

### DIFF
--- a/src/resdk/resources/geneset.py
+++ b/src/resdk/resources/geneset.py
@@ -1,0 +1,171 @@
+"""Geneset resource."""
+import json
+import logging
+from urllib.parse import urljoin
+
+from .data import Data
+from .utils import get_collection_id
+
+
+class Geneset(Data):
+    """Resolwe Geneset resource.
+
+    :param resolwe: Resolwe instance
+    :type resolwe: Resolwe object
+    :param model_data: Resource model data
+
+    """
+
+    def __init__(self, resolwe, genes=None, source=None, species=None, **model_data):
+        """Initialize attributes."""
+        self.logger = logging.getLogger(__name__)
+
+        self._genes = None
+        # Make sure genes are stored in a set object
+        if genes is not None:
+            self._genes = set(genes)
+
+        self._source = source
+        self._species = species
+
+        super().__init__(resolwe, **model_data)
+
+    @property
+    def genes(self):
+        """Get genes."""
+        if self._genes is None or len(self._genes) == 0:
+            if self.id and "geneset_json" in self.output:
+                url = urljoin(
+                    self.resolwe.url,
+                    "api/storage/{}".format(self.output["geneset_json"]),
+                )
+                response = self.resolwe.session.get(url, auth=self.resolwe.auth)
+                response = json.loads(response.content.decode("utf-8"))
+                self._genes = set(response["json"]["genes"])
+        return self._genes
+
+    @genes.setter
+    def genes(self, genes):
+        """Set genes."""
+        self._assert_allow_change("genes")
+        if genes is not None:
+            self._genes = set(genes)
+
+    @property
+    def source(self):
+        """Get source."""
+        if self._source is None and self.id and "source" in self.output:
+            self._source = self.output["source"]
+        return self._source
+
+    @source.setter
+    def source(self, new_source):
+        """Set source."""
+        self._assert_allow_change("source")
+        self._source = new_source
+
+    @property
+    def species(self):
+        """Get species."""
+        if self._species is None and self.id and "species" in self.output:
+            self._species = self.output["species"]
+        return self._species
+
+    @species.setter
+    def species(self, new_species):
+        """Set species."""
+        self._assert_allow_change("species")
+        self._species = new_species
+
+    def _assert_allow_change(self, field_name):
+        """Assert that this Geneset obj is not saved yet."""
+        if self.id:
+            msg = "Not allowed to change field {} after geneset is saved".format(
+                field_name
+            )
+            raise ValueError(msg)
+
+    def save(self):
+        """Save Geneset to the server.
+
+        If Geneset is already on the server update with save() from base class. Otherwise, create
+        a new Geneset by running process with slug "create-geneset".
+        """
+
+        if self.id:
+            super().save()
+        else:
+            none_fields = [
+                name
+                for name in ["genes", "source", "species"]
+                if getattr(self, name, None) is None
+            ]
+
+            if none_fields:
+                msg = "Fields {} must not be none".format(", ".join(none_fields))
+                raise ValueError(msg)
+
+            data = {
+                "process": {"slug": "create-geneset"},
+                "input": {
+                    "genes": list(self.genes),
+                    "source": self.source,
+                    "species": self.species,
+                },
+            }
+            if self.name:
+                data["name"] = self.name
+            if self.collection:
+                data["collection"] = {"id": get_collection_id(self.collection)}
+
+            model_data = self.api.post(data)
+            tmp_genes, tmp_source, tmp_species = self.genes, self.source, self.species
+            self._update_fields(model_data)
+            # Since there is no output values in model_data
+            # the original genes, source and species values gets overwritten
+            # so we set them back here
+            self._genes, self._source, self._species = (
+                tmp_genes,
+                tmp_source,
+                tmp_species,
+            )
+
+    def __and__(self, other):
+        """Intersection."""
+        return self.set_operator("__and__", other)
+
+    def __or__(self, other):
+        """Union."""
+        return self.set_operator("__or__", other)
+
+    def __sub__(self, other):
+        """Difference."""
+        return self.set_operator("__sub__", other)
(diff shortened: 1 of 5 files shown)


src/resdk/resources/geneset.py
```diff
@@ -0,0 +1,118 @@
...
+        self.logger = logging.getLogger(__name__)
+
+        self._genes = None
+        self.source = None
+        self.species = None
+
+        super().__init__(resolwe, **model_data)
+
+    @property
+    def genes(self):
```
JureZmrzlikar, line 70: This should go on the server and get the genes. Use this code to fetch them:
```
url = urljoin(self.resolwe.url, "api", "storage", self.output["geneset_json"])
response =  self.resolwe.session.get(url)
genes = json.loads(response.content)["json"]["genes"]
```

src/resdk/resources/geneset.py
```diff
@@ -0,0 +1,118 @@
...
+        "source",
+        "species",
+    )
+
+    def __init__(self, resolwe, **model_data):
+        """Initialize attributes."""
+        self.logger = logging.getLogger(__name__)
+
+        self._genes = None
+        self.source = None
```
JureZmrzlikar, line 64: Aster some thinking i believe it would be best to also implement source and species as getter/setter methods. In case geneset is created from scratch, locally these method set the `self._species/_source` attributes. In case it is pulled from the server, the values for species/source are pulled from `Geneset` (`Data`) outputs: `self.output["source/species"]`

How do you differ between Geneset was/is pulled from the server vs. created locally? By checking the `self.id` attribute. Similar approach is used here:
https://github.com/genialis/resolwe-bio-py/blob/master/src/resdk/resources/base.py#L124

tests/unit/test_geneset.py
```diff
@@ -0,0 +1,80 @@
...
+        for gene in genes:
+            self.assertTrue(gene in gs.genes)
+
+    def test_arguments(self):
+        gs = Geneset(
+            MagicMock(), genes=["FHIT", "MYC"], source="UCSC", species="Homo sapiens"
+        )
+
+        # test all geneset attributes
+        self._check_equal_gs(gs, ["FHIT", "MYC"])
```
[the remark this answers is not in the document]
AGregorc: Thanks for mentioning this function. I used it now and deleted my helper function.

src/resdk/resources/geneset.py
```diff
@@ -0,0 +1,118 @@
...
+
+        :param genes: list
+        """
+        self.genes = []
+        if genes is not None:
+            for value in genes:
+                if value not in self.genes:
+                    self.genes.append(value)
+
+    def __iter__(self):
```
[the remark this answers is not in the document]
AGregorc: I think these are the only methods you need to implement if you're subclassing `Set`. But this class is deleted now anyway :)

src/resdk/resources/geneset.py
AGregorc: I used that code now for fetching genes. I hope it is done the right way 😄

src/resdk/resources/geneset.py
AGregorc: Thanks for that idea. I implemented getter/setter methods for source and species.

With the last commit `gs = res.geneset.create(genes=[...], species=..., source=...)` should work. 
I updated some unit tests but it would probably be a good idea to add some functional tests as well.

tests/unit/test_geneset.py
```diff
@@ -0,0 +1,98 @@
...
+        base_mock.api = MagicMock(**{"patch.return_value": {"id": 1, "slug": "test"}})
+        Geneset.save(base_mock)
+        self.assertEqual(base_mock._update_fields.call_count, 1)
+
+    def test_create(self):
+        gs = Geneset(
+            MagicMock(), genes=["FHIT", "MYC"], source="UCSC", species="Homo sapiens"
+        )
+        gs.api = MagicMock(**{"post.return_value": {"id": 1, "slug": "test"}})
+        gs.save()
```
JureZmrzlikar, line 89: These two lines can actually be deleted and test will still pass :-)
I know that it can be challenging to write a good test with so much mocking. Consider just checking that post request was made with expected payload.

tests/unit/test_geneset.py
```diff
@@ -0,0 +1,98 @@
...
+"""
+
+import unittest
+
+from mock import MagicMock, patch
+
+from resdk.resources.geneset import Geneset
+
+
+class TestData(unittest.TestCase):
```
JureZmrzlikar, line 12: TestGeneset

AGregorc:
Now the genes, source and species variables are not part of a base class fields and all tests are finally passing.

src/resdk/resources/geneset.py
```diff
@@ -0,0 +1,161 @@
...
+    def save(self):
+        """Save Geneset to the server.
+
+        If Geneset is already on the server update with save() from base class. Otherwise, create
+        a new Geneset by running process with slug "create-geneset".
+        """
+
+        if self.id:
+            super().save()
+        else:
```
JureZmrzlikar, line 97: Since the new constructor can accept values of source species and genes as `None`, add a check that verifies that these values are not `None` at the time of Geneset creation.

src/resdk/resources/geneset.py
```diff
@@ -0,0 +1,142 @@
...
+                    "api/storage/{}".format(self.output["geneset_json"]),
+                )
+                response = self.resolwe.session.get(url, auth=self.resolwe.auth)
+                response = json.loads(response.content.decode("utf-8"))
+                self._genes = set(response["json"]["genes"])
+        # Convert set to list since object of type set is not JSON serializable.
+        return list(self._genes)
+
+    @genes.setter
+    def genes(self, genes_list):
```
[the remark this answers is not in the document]
JureZmrzlikar: Ping.

src/resdk/resources/geneset.py
JureZmrzlikar: Otherwise process will fail.

tests/functional/genesets/e2e_genesets.py
```diff
@@ -0,0 +1,38 @@
...
+    def tearDown(self):
+        # self.geneset is deleted along with collection
+        self.collection.delete(force=True)
+
+    def test_create_geneset(self):
+        self.geneset = self.res.geneset.create(
+            name="Test name",
+            genes=["FHIT", "MYC"],
+            source="UCSC",
+            species="Homo sapiens",
```
JureZmrzlikar, line 21: I would prefer to chck the "full" version of create, including collection argument.

tests/functional/genesets/e2e_genesets.py
```diff
@@ -0,0 +1,38 @@
...
+
+        self.assertTrue(isinstance(self.geneset, resdk.resources.Geneset))
+        self.assertCountEqual(self.geneset.genes, ["FHIT", "MYC"])
+        self.assertEqual(self.geneset.source, "UCSC")
+        self.assertEqual(self.geneset.species, "Homo sapiens")
+        self.assertEqual(self.geneset.name, "Test name")
+
+        # Update geneset with a new collection and tag
+        self.geneset.collection = self.collection
+        self.geneset.collection.tags = ["community:expressions"]
```
JureZmrzlikar, line 32: If you include collection argument in the `create` method, you can leave out collection here. In that case, assign tag to the geneset, not the collection.

JureZmrzlikar (commented):
A few smaller fixes, but this is almost done. Good work!
After these changes, remove the "WIP" prefix of the PR title and squash the commits in one.

src/resdk/resources/geneset.py
```diff
@@ -0,0 +1,170 @@
...
+        """Save Geneset to the server.
+
+        If Geneset is already on the server update with save() from base class. Otherwise, create
+        a new Geneset by running process with slug "create-geneset".
+        """
+
+        if self.id:
+            super().save()
+        else:
+            none_fields = []
```
JureZmrzlikar, line 98: Could also be in more compact form:
none_fields = [name for name in ["genes", "source", "species"] if getattr(self, name) is None]

src/resdk/resources/geneset.py
```diff
@@ -0,0 +1,170 @@
...
+            for name in ["genes", "source", "species"]:
+                if getattr(self, name, None) is None:
+                    none_fields.append(name)
+
+            if len(none_fields) > 0:
```
JureZmrzlikar, line 103: It is sufficient to say `if none_fields:`

JureZmrzlikar (approved):
Nice, just a few stylistic changes. 
As said, please demove pr WIP prefix of this PR and merge commits in one.

PR #251, opened 2021-08-10
Fix misordered meta and expression data index

### LINKED ISSUE #250: CollectionTables returns expressions and metadata with different order of samples
Indices of `ct.exp` and `ct.meta`, where `ct` is a `resdk.collection_tables.CollectionTables` object, are in general not the same. The problem occurs when expressions and the associated response, which is typically stored in the metadata, are used as input in a `sklearn` estimator. Unfortunately, the `fit` method ignores the index of `pandas.DataFrame` and `pandas.Series` objects and matches expression profiles with response data by integer index and not by sample name. Thus, if the order of samples in `ct.exp` and `ct.meta` is different, reshuffled (wrong) responses may be assigned to the samples.

@robertcv proposed a solution that computes the intersection of expression and metadata indices and returns both `pandas.DataFrame`s with the same order of rows, removing samples with no expression data objects.

The current workaround is to sort expressions and metadata by their sample index in each script after obtaining them from the server.

robertcv:
Fixes #250 
Somewhere along the way of improving and extending CollectionTables functionality, we lost the sorting of indexes in metadata which makes `ct.exp` and `ct.meta` have different ordered samples.

### DIFF
--- a/src/resdk/collection_tables.py
+++ b/src/resdk/collection_tables.py
@@ -502,6 +502,7 @@ class CollectionTables:
             how = "right" if meta.columns.empty else "outer"
             meta = meta.merge(orange_data, how=how, left_index=True, right_index=True)
 
+        meta = meta.sort_index()
         meta.index.name = "sample_name"
 
         return meta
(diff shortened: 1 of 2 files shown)


robertcv:
@JureZmrzlikar It would be great if we could get a hotfix release for this.

JureZmrzlikar:
Looks good, thanks for the fix! Please rebase on the latest master.

JureZmrzlikar (approved):

PR #258, opened 2021-08-20
WIP: Generalize CollectionTables

JureZmrzlikar:


JureZmrzlikar:
@mstajdohar @robertcv 
Please review conceptually, i'd like some feedback before polishing the code.

mstajdohar:
Let's discuss live when you return.

JureZmrzlikar:
@mstajdohar @robertcv 
I might add a test or two but in general my work here is done. Please review.

src/resdk/collection_tables.py
```diff
@@ -1,632 +1,16 @@
...
-        the needed subset of fields is fetched.
-
-        :return: list od Sample objects
-        """
-        sample_ids = [d.sample.id for d in self._data]
-        return list(
-            self.collection.samples.filter(id__in=sample_ids, fields=SAMPLE_FIELDS)
+        """Initialize."""
+        warnings.warn(
+            "resdk.CollectionTables will be deprecated in the next major "
```
mstajdohar, line 13: We change major versions often. Should we report a timestamp instead? E.g., in 2022Q1.

src/resdk/tables/__init__.py
```diff
@@ -0,0 +1,23 @@
+""".. Ignore pydocstyle D400.
+
+============
+ReSDK Tables
+============
+
```
mstajdohar, line 6: Provide a brief description of the `tables` module. For example "Helper classes for aggregating collection data in ta tabular format. "

mstajdohar (approved):
LGTM

src/resdk/collection_tables.py
JureZmrzlikar: Good suggestion, thanks!

PR #259, opened 2021-08-26
Speedup of merging individual sample expressions

robertcv:
Concatenating pandas `Series` is faster and uses less RAM than single raw `DataFram` objects. This is especially noticeable for manually uploaded expressions that don't contain all genes.


JureZmrzlikar (approved):
All looks good, but could you please rename `df` and `df_list` to something "Series-like"? It's knida confusing if a `pd.Series` (or a list of ti) is named `df*`...

JureZmrzlikar (approved):

PR #269, opened 2021-11-08
Support retrieval of QC values in ReSDK tables

JureZmrzlikar:


jkokosar (approved):
I focused on the MQC_GENERAL_COLUMNS data (column names, agg_function), and all looks good to me. I've tested this functionality using different collections with real data, and haven't found any issues. 

@gregorjerse please do a more technical review.

src/resdk/tables/base.py
```diff
@@ -213,6 +223,30 @@ def _metadata_version(self) -> str:
...
+    def _qc_version(self) -> str:
+        """Return server QC data version.
+
+        The versioning of QC Data on the server is determined by the hash of
+        the tuple of sorted QC data objects ids.
+
+        :return: data version
+        """
+        mqc_ids = []
+        for mqc in self.collection.data.filter(
```
gregorjerse, line 237: Would not the same be achieved by 

``mqc_ids = self.collection.data.filter(type="data:multiqc", status="OK", fields=["id", "entity__id"], entity__isnull=True)``  ?

src/resdk/tables/base.py
```diff
@@ -213,6 +223,30 @@ def _metadata_version(self) -> str:
...
+            type="data:multiqc", status="OK", fields=["id", "entity__id"]
+        ):
+            if mqc.sample is None:
+                continue
+            mqc_ids.append(mqc.id)
+        if len(mqc_ids) == 0:
```
gregorjerse, line 243: It is probably more `Pythonic` to simplify the condition to  ``if mqs_ids:``

src/resdk/tables/base.py
```diff
@@ -213,6 +223,30 @@ def _metadata_version(self) -> str:
...
+            raise ValueError(
+                f"Collection {self.collection.name} has no samples with MultiQC data!"
+            )
+
+        return str(hash(tuple(sorted(mqc_ids))))
```
gregorjerse, line 248: If you add ``ordering="id"`` to the filter when reading `mqs_ids` you can get rid of the `sorted` call.

src/resdk/tables/rna.py
```diff
@@ -21,10 +21,196 @@
...
+
+
+def multiqc_general_stats_parser(file_object, name):
+    """Parse "multiqc_general_stats.txt" file."""
+    df = pd.read_csv(file_object, sep="\t", index_col=0)
+
+    # Keep only specified columns:
+    df = df[
+        [
+            c.get("name", "")
```
gregorjerse, line 168: I know it is tempting to use short variable names, but I prefer more descriptive ones.
Such as `data_frame` instead of `df`, `column` instead of `c`, `aggregate??` instead of `s` ...

I believe it makes the code more readable.

src/resdk/tables/rna.py
```diff
@@ -216,12 +402,61 @@ def _cache_file(self, data_type: str) -> str:
...
+        """
+        mqc_db = {}
+        for mqc in self.collection.data.filter(
+            type="data:multiqc",
+            status="OK",
+            ordering="created",
+            fields=["id", "name", "entity__id"],
+        ):
+            if mqc.sample is None:
+                continue
```
gregorjerse, line 427: Same comment as above, this if could be transformed to a filter.

src/resdk/tables/rna.py
```diff
@@ -216,12 +402,61 @@ def _cache_file(self, data_type: str) -> str:
...
+            mqc_db[mqc.sample.id] = {
+                "mqc_id": mqc.id,
```
gregorjerse, line 429: Is it possible to have more than one `mqc` data object per sample? Probably not, but without this assumption the code does not work.

src/resdk/tables/rna.py
```diff
@@ -216,12 +402,61 @@ def _cache_file(self, data_type: str) -> str:
...
+                "uri_general": f"{mqc.id}/multiqc_data/multiqc_general_stats.txt",
+                "uri_strand": f"{mqc.id}/multiqc_data/multiqc_library_strandedness.txt",
+                "uri_build": f"{mqc.id}/multiqc_data/multiqc_sample_info.txt",
+            }
+
+        df = pd.DataFrame(index=[s.id for s in self._samples])
```
gregorjerse, line 435: More descriptive variable names, please.

src/resdk/tables/rna.py
```diff
@@ -216,12 +402,61 @@ def _cache_file(self, data_type: str) -> str:
...
+
+        uris = [item["uri_general"] for item in mqc_db.values()]
```
gregorjerse, line 437: This entire range could be replaced with something like

```
parsers = {"uri_general": multiqc_general_stats_parser, "uri_strand": ..., "uri_build": ...}
for type_, parser in parsers.items():
    uris = [item[type_] for item in mqc_db.values()]
    qc = batch_download(self.resolwe, uris, parser)
    qc = qc.rename(index={value[type_]: key for key, value in mqc_db.items()})
    df = df.merge(qc, how="left", left_index=True, right_index=True)
```

gregorjerse (approved):

src/resdk/tables/rna.py
jkokosar: The platform will not prevent you from re-running the MutiQC analysis on the same sample data, so it is possible to have multiple objects on the same sample. I suggest using the latest one in such cases, and reporting the warning to the user.

src/resdk/tables/base.py
JureZmrzlikar: Good point, thanks.

src/resdk/tables/base.py
JureZmrzlikar: Good point, thanks.

src/resdk/tables/base.py
JureZmrzlikar: Good point, thanks.

src/resdk/tables/rna.py
JureZmrzlikar: I've done this for all but `df`, which is IMO the community accepted notation for a `pandas.DataFrame`.

src/resdk/tables/rna.py
JureZmrzlikar: Yes, this was exactly the reason why it was made as is. If there are multiple MultiQC Data in one Sample, only the latest will be used.

PR #271, opened 2021-11-19
Add resdk.tables.MATables for microarray data support

JureZmrzlikar:


src/resdk/tables/microarray.py
```diff
@@ -0,0 +1,108 @@
...
+
+        # Fetch collection microarray and metadata
+        tables = MATables(collection)
+        meta = tables.meta
+        exp = tables.exp
+
+    """
+
+    process_type = "data:microarray:normalized"
+    EXP = "exp"
```
robertcv, line 48: I would suggest using some other label like `"ma"`. Although the cached file name will not be the same as the hash of data objects is different I would still like it to be different so it's more obvious which cache file contains what.

src/resdk/tables/microarray.py
```diff
@@ -0,0 +1,108 @@
+""".. Ignore pydocstyle D400.
+
+========
```
robertcv, line 3: You should also add some lines to the topic documentation.

robertcv (changes requested):

src/resdk/tables/microarray.py
JureZmrzlikar: Good point, thanks!

src/resdk/tables/microarray.py
JureZmrzlikar: Done.

JureZmrzlikar:
@robertcv I've addressed your comments, please re-review.

robertcv (changes requested):
The changes look OK. I just noticed one more thing. I would like it for exp to include the `attrs` attribute similar to RNATables. I would add parameters `exp_type` and `platform` from the data object output attributes.

robertcv (approved):

gregorjerse (approved):

PR #281, opened 2022-03-11
Add Metadata resource

JureZmrzlikar:


JureZmrzlikar:
Hi @mzganec @robertcv @AnzeLovse 

IMO the code in `resdk.resources.Metadata` is mature enough for review. The main thing that is missing are unit tests and that is why I'm leaving this as WIP. To ease the start, you can also check the docs section with `metadata.rst` or my personal hackish e2e test in `test-metadata.py` (in root dir). This file is also the reason why packaging tests fail and will be removed before merging.

I would like general feedback on the design of code and weather it meets you expectations (especially @mzganec  here). Fell fee to also drop your comments on other stuff, but IMO this should be discussed first.

I intentionally left out the setting of DataFrame index to sample ID / slug / name column.

src/resdk/resources/metadata.py
```diff
@@ -0,0 +1,232 @@
...
+            if self.df is None or self.df.empty:
+                raise ValueError("Attribute df must be set before saving.")
+            if not self.collection:
+                raise ValueError("Collection must be set before saving.")
+
+            # All resdk machinery for uploading files works with real
+            # files on the system. Ideally we would support "file upload"
+            # from a stream, but for now, let's use a tempfile solution.
+            tmp_file = tempfile.NamedTemporaryFile(suffix=".csv")
+            self.df.to_csv(tmp_file.name)
```
AnzeLovse, line 203: Would it make sense to skip writing the index here, since DataFrame index is not being set to meaningful values?

src/resdk/resources/metadata.py
```diff
@@ -0,0 +1,232 @@
...
+            # Otherwise try to guess the parser based on file extension
+            basename = self.output["table"]["file"]
+            if parser is not None:
+                df = parser(self.df_bytes, **kwargs)
+            elif basename.endswith("xls"):
+                df = pd.read_excel(self.df_bytes, engine="xlrd")
+            elif basename.endswith("xlsx"):
+                df = pd.read_excel(self.df_bytes, engine="openpyxl")
+            elif any(basename.endswith(ext) for ext in ["tab", "tsv"]):
+                df = pd.read_csv(self.df_bytes, sep="\t")
```
AnzeLovse, line 160: We should be mindful of float precision and rounding when loading data; [see this example](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#specifying-method-for-floating-point-conversion). Say we get the data frame from collection, do some modifications and save it afterward. This could introduce undesired changes to the metadata. I am not sure how often this can happen but users should at least be made aware of it.

AnzeLovse (commented):
I have no particular design comments but would be interested in how will this work with `Tables.meta`. Will they become obsolete? Is it possible to merge some of the code? Overall, it looks like a neat functionality to simplify metadata handling.

src/resdk/resources/metadata.py
JureZmrzlikar: Thanks for pointing this out, added "round_trip" option for float precision.

JureZmrzlikar:
> would be interested in how will this work with Tables.meta. Will they become obsolete? Is it possible to merge some of the code?

Good point. For now, I envision they will be separate, even if they contain some duplicated functionality. I would like to enable  Metadata to become more mature and develop in whichever way it will benefit users most. If the two remain very close I am in favor of merging the functionality. But lets wait and see.

tests/unit/test_metadata.py
```diff
@@ -0,0 +1,137 @@
...
+    def test_init(self):
+        meta = Metadata(self.res, df=self.df)
+
+        self.assertEqual(meta.resolwe, self.res)
+        pd.testing.assert_frame_equal(meta.df, self.df)
+
+    def test_unique(self):
+        # If created from scratch, unique should be True by default
+        meta = Metadata(self.res)
+        # TODO: Why is this false?
```
robertcv, line 55: In the `Metadata.__init__` you set `self.unique` to `True` and thereby call `self.unique` setter method which uses the `self.resolwe` which is a `Mock` to set the `self.process` which then also is a `Mock` and in the unique getter you use that Mock for testing which then files. You should rewrite the test or extend the `self.res` `Mock` in the `setUp`

test-metadata.py
```diff
@@ -0,0 +1,40 @@
+"""Test script for resdk.resources.Metadata."""
```
robertcv, line 1: This is great to quickly test the new functionality but don't forget to remove it before merging.

src/resdk/resources/metadata.py
```diff
@@ -0,0 +1,247 @@
...
+            )
+
+        # Sample identifiers from df
+        # Convert to string to avoid string vs. integer comparisons
+        df_samples = set(df[match_column].astype(str).tolist())
+
+        # Sample identifiers from collection
+        # Convert to string to avoid string vs. integer comparisons
+        attr = self.sample_identifier_columns[match_column]
+        col_samples = {str(getattr(s, attr)) for s in self.collection.samples}
```
robertcv, line 123: You should add a filter with fields set to only attr to improve speed.

docs/metadata.rst
```diff
@@ -0,0 +1,81 @@
...
+
+In the most common case, Metadata objects exist somewhere on Resolwe
+server and user just fetches them::
+
+    # Get one metadata by slug
+    m = res.metadata.get("my-slug")
+
+    # Filter metadata by some conditions, e.g. get all metadata
+    # from a given collection:
+    ms = res.metadata.filter(collection=<my-collection>):
```
robertcv, line 51: ms -> m

robertcv (commented):
Overall looks ok. The only bigger change I would suggest is perhaps using df index as the sample identifier. So when calling set_ds it would expect the identifier as the index and when calling get_df it would set the identifier column as the index.

src/resdk/resources/metadata.py
```diff
@@ -0,0 +1,247 @@
...
+            if self.df is None or self.df.empty:
+                raise ValueError("Attribute df must be set before saving.")
+
+            # All resdk machinery for uploading files works with real
+            # files on the system. Ideally we would support "file upload"
+            # from a stream, but for now, let's use a tempfile solution.
+            tmp_file = tempfile.NamedTemporaryFile(suffix=".csv")
+            self.df.to_csv(tmp_file.name, index=False)
+            inputs = self.resolwe._process_inputs({"src": tmp_file.name}, self.process)
+            tmp_file.close()
```
mzganec, line 220: @robertcv Can you weigh in here? In the first line, a temporary file is created and opened for writing. In the next line, the same file is opened again for writing by `pandas` using a different handler. Does this work on all operating systems, i.e., no locking of files opened for writing? After writing, `pandas`  closes the file, which is where the flush occurs as well, so we have a guarantee to have the full file content available for upload in the next line. If we now wrote something to the same file with `tmp_file.write(b"blabla")` and flushed, the written `DataFrame` would be partially overwritten by `"blabla"`. After `self.df.to_csv(...)` the temporary file is not deleted yet even though the file has been closed by `pandas` because the deletion of temporary files is handled by `tmp_file.close()`.

mzganec (approved):
I ran `tests/unit/test_metadata.py` and `test-metadata.py` on macOS and everything works well.

src/resdk/resources/metadata.py
mzganec: @JureZmrzlikar Perhaps `self.df.to_csv(tpm_file, index=False)` would be a better choice to avoid creating the second file handler and reopening the same file for writing, but I'm not sure if `pandas` flushes afterwards.

docs/metadata.rst
JureZmrzlikar: Filter returns multile elements, so IMO, ms (as a plural of m) is OK.

tests/unit/test_metadata.py
JureZmrzlikar: Thanks.

src/resdk/resources/metadata.py
JureZmrzlikar: Flush is now called manually to ensure all content is present in the temporary file.

AnzeLovse (approved):
End to end tests also pass OK on mac. LGTM

test-metadata.py
JureZmrzlikar: Removed.

Issue #284, opened 2022-04-06
Table.meta not working with the latest upload-metadata processes


AGregorc:
Since the Sample ID/slug/name naming remove the orange type `mS#...` in the new processes, `Table.meta` doesn't work anymore. 
The problem lies from this [line onward](https://github.com/genialis/resolwe-bio-py/blob/master/src/resdk/tables/base.py#L405). 

Example to reproduce error:
```python
slug = <any_new_collection_with_new_metadata_process>

res = resdk.Resolwe(url="https://app.genialis.com/")
table = RNATables(res.collection.get(slug))
table.meta
```

JureZmrzlikar:
Fixed by #285

PR #285, opened 2022-04-06
Fix parsing new metadata format in resdk.tables.BaseTables.meta

JureZmrzlikar:


JureZmrzlikar:
Fixes #284

AGregorc (approved):
lgtm

gregorjerse (approved):

src/resdk/tables/base.py
```diff
@@ -402,17 +402,31 @@ def _get_orange_data(self) -> pd.DataFrame:
                 # TODO: logging, warning?
                 return pd.DataFrame()
 
-        if "mS#Sample ID" in df.columns:
+        if "Sample ID" in df.columns:
```
mstajdohar, line 405: How about making this even more general? For example,
```
for key in df.columns:
    if "sample id" in key.lower()
        ...
```

JureZmrzlikar: This PR is merged already. Valid point, i'll add this to one of the future PR's.

PR #288, opened 2022-04-13
Metadata: Add column if match is in df index, remove matched column

AGregorc:
While I was using Metadata resource I got some ideas how this resource could be improved. In this PR I added three things:

- Usually, `Sample name/slug/ID` is already set as index and `m.set_index(<df>)` could just map it to `Sample ID`.
- `m.set_index(<df>)` could also delete `Sample name/slug/ID` column since it is not needed anymore. The same as `DataFrame.set_index` function removes the column.
- Add example with `m.set_index(<df>)` function in the documentation.

These are generally just ideas and if somebody else can find them useful then I'll be glad if we merge this PR.

### DIFF
--- a/src/resdk/resources/metadata.py
+++ b/src/resdk/resources/metadata.py
@@ -89,9 +89,19 @@ class Metadata(Data):
         return self._df_bytes
 
     def set_index(self, df):
-        """Set index of df to Sample ID."""
+        """Set index of df to Sample ID.
+
+        If there is a column with ``Sample ID`` just set that as index. If there is
+        ``Sample name`` or ``Sample slug`` column, map sample name / slug to sample ID's
+        and set ID's as an index. If no suitable column in there, raise an error.
+        Works also if any of the above options is already an index with appropriate name.
+        """
         for match_column in self.sample_identifier_columns:
             if match_column in df.columns:
+                break
+            if match_column == df.index.name:
+                # Add new column with index name
+                df[match_column] = df.index
                 break
         else:
             options = ", ".join(self.sample_identifier_columns)
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -17,6 +17,9 @@ Fixed
 -----
 - Fix parsing of new metadata format in ``resdk.tables.BaseTables.meta``
 
+Changed
+-------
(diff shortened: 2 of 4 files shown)


src/resdk/resources/metadata.py
```diff
@@ -117,6 +121,9 @@ def set_index(self, df):
         # Remove the samples that do not have mapping
         df = df.dropna(subset=["Sample ID"])
 
+        # We don't need the sample identifier column anymore
```
JureZmrzlikar, line 124: I would skip this - the removal part. I can understand that you do not need this column anymore in your current use cases, and the similarity with `DataFrame.set_index()` method behaviour. However, IMO i't a bit of an overkill to remove e.g. "Sample name" column... For some use cases it might come handy. I would rather keep it and let users remove them themselves if necessary that having to to some jujitsu to keep it in there.

src/resdk/resources/metadata.py
```diff
@@ -93,6 +93,10 @@ def set_index(self, df):
         for match_column in self.sample_identifier_columns:
```
JureZmrzlikar, line 93: Would you mind also expanding the docstring of this method a bit more? The name is same as for `DataFrame.set_index()`, but what happens behind the scenes here is much more. It was my mistake to not do this in the first place, but since you're actually showcasing this method in the docs it's probably the right time to do it now.

I would go with something like:

Set index of `df` to sample ID.

If there is a column with Sample ID just set that as index. If there is a column with sample slug or name, map sample name / slug to sample ID's and set ID's as an index. If no suitable column in there, raise an error. Works also if any of the above options is already an index with appropriate name.

JureZmrzlikar (commented):
Nice suggestion, I like it, especially if it keeps dowstream code simpler. Please also add a changelog entry.

JureZmrzlikar (commented):
IMO the docs (`docs/metadata.rst`) file was appropriate in the initial commit, but I want you to expand the docstring of the method `Metadata.set_index()`. Sorry if this was not clear from my review.
Otherwise this looks good to merge, thanks.

AGregorc:
Thanks for the review. I updated the code and squashed everything to one commit.

JureZmrzlikar (approved):

PR #289, opened 2022-04-13
Variant tables

JureZmrzlikar:


src/resdk/tables/variant.py
```diff
@@ -0,0 +1,157 @@
...
+        VARIANTS: "tsv",
+        DEPTH: "tsv",
+        COVERAGE: "tsv",
+    }
+
+    @property
+    @lru_cache()
+    def mutation_subset_genes(self) -> pd.DataFrame:
+        """Return genes for which variants are reported."""
+        return self._data[0].output["genes"]
```
AnzeLovse, line 54: The docstring could be extended because it could be misleading for someone not familiar with `mutations-table`. For example, we use gene `FOO` and its mutation `Val10` as an input to the `mutations-table`. After using the function and getting `["FOO"]` as a list of genes one could naively assume that all other mutations of this gene were considered as well.

src/resdk/tables/variant.py
```diff
@@ -0,0 +1,157 @@
...
+
+        s.name = sample_id
+        return s
+
+    async def _download_data(self, data_type: str) -> pd.DataFrame:
+        df = await super()._download_data(data_type)
+        # If all samples are missing a variant, remove it
+        df = df.dropna(axis=1, how="all")
+        # No mutation detected means WT, ergo encode 0
+        return df.fillna(0)
```
AnzeLovse, line 157: What happens if we use different genes for different samples?

If we had `mutation-tables` for one gene on one sample and another gene one the other sample wouldn't `df.fillna(0)` transform non-analysed genes to a WT.

AnzeLovse (commented):
Overall this looks great! The tests pass and I successfully get the results from the toy collection.

src/resdk/tables/variant.py
```diff
@@ -0,0 +1,157 @@
...
+        # No QC is given for variants data - return empty DataFrame
+        return pd.DataFrame()
+
+    @staticmethod
+    def construct_index(row) -> str:
+        """
+        Construct index of the variants table.
+
+        Index should have the form:
+        <chr>_<position>_<snp_change>_<aminoacid_pos_change>
```
OtonicarJan, line 85: It would be great if the gene name is also added to index if we run the analysis for multiple genes at once.

src/resdk/tables/variant.py
```diff
@@ -0,0 +1,157 @@
...
+            self.construct_index, axis=1, result_type="reduce"
+        )
+        sample_data.set_index("index", inplace=True)
+
+        if data_type == self.VARIANTS:
+            s = sample_data.apply(self.encode_mutation, axis=1, result_type="reduce")
+        elif data_type == self.DEPTH:
+            # Depth, as computed by GATK is reported by "DP" column
+            s = sample_data["DP"]
+        elif data_type == self.COVERAGE:
```
OtonicarJan, line 145: When I ran the test in the toy collection, I get this output: 
```
>>> vt.coverage
index      chr12_25245350_C>T_Gly12Asp
sample_id                             
9073                               0.0
9075                               91.0
```
The coverage for the sample 9073 should be reported (96). Coverage ("Total_depth") is always in the output in mutations table, so it should also be in the output here even if there is no mutation present.

src/resdk/tables/variant.py
```diff
@@ -0,0 +1,157 @@
...
+        # Fetch collection variant data
+        tables = VariantTables(collection)
+        var = tables.variants
+
+    """
+
+    process_type = "data:mutationstable"
+    VARIANTS = "variants"
+    DEPTH = "depth"
+    COVERAGE = "coverage"
```
OtonicarJan, line 42: Reporting coverage is OK, but reporting base frequencies would offer even more insight about the variant. If we indeed report base frequencies then maybe coverage report can be removed.

OtonicarJan (commented):
Looks good. Left few comments.

src/resdk/tables/variant.py
```diff
@@ -0,0 +1,157 @@
...
+        sample_data.dropna(axis=0, subset=["ALT"], inplace=True)
+
+        # Construct index
+        sample_data["index"] = sample_data.apply(
+            self.construct_index, axis=1, result_type="reduce"
+        )
+        sample_data.set_index("index", inplace=True)
+
+        if data_type == self.VARIANTS:
+            s = sample_data.apply(self.encode_mutation, axis=1, result_type="reduce")
```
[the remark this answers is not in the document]
JureZmrzlikar: This happens because there are `NaN` values present for this particluar variant. NaN's are floats, so all values in a given column are casted to float. I would rather keep the possiblity of having `np.nan`'s than to cast this to integers and present missing values with pandas.NA (which can represent missing integers).

src/resdk/tables/variant.py
```diff
@@ -0,0 +1,157 @@
...
+            - 1 for heterozygous mutation
+            - 2 for homozygous mutation
+        """
+        try:
+            allele1, allele2 = re.match(
+                r"([ACGT])/([ACGT])", row["SAMPLENAME1.GT"]
+            ).group(1, 2)
+        except (AttributeError, TypeError):
+            # TypeError is raised when row["SAMPLENAME1.GT"] is np.nan / None
+            # AttributeError is raised when there is no match
```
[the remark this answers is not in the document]
JureZmrzlikar: I don't really know. I wanted to handle the case of empty string (`""`), but we might want to reconsider this.

src/resdk/tables/variant.py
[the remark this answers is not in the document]
AnzeLovse: I agree, keeping in the `NaN` values makes sense and is more informative.

src/resdk/tables/variant.py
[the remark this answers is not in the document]
AnzeLovse: It is nice to handle that as well. Perhaps returning `np.nan` would be more appropriate in the case of an empty string.

JureZmrzlikar:
Hi @AnzeLovse and @OtonicarJan , I'm asking for (at least) another review pass.

After trying to apply your fixes i saw the need to optimize / generalize things a bit more... So here you have. IMO I've addressed your comments but the amount of changes looks more like a rewrite. Feel free to ask questions. Or just point our parts that are unclear - we probably want to document those in more depth. Maybe a good place to start reviewing is the docs part to get the general "why" behind some of the new functionality.

AnzeLovse (approved):
Amazing effort! The expended documentation is great and the code seems more robust now. Remember to merge WIP commits.

src/resdk/tables/variant.py
```diff
@@ -0,0 +1,424 @@
...
+        data = []
+        sample_ids, repeated_sample_ids = set(), set()
+        for datum in self.collection.data.filter(
+            type=self.process_type,
+            status="OK",
+            ordering="-created",
+            fields=self.DATA_FIELDS,
+        ):
+            # 1 Filter by mutations, if required
+            if self.mutations:
```
OtonicarJan, line 178: Currently, if you call `VariantTables` and set a certain "gene: aminoacid" combination that is not present in the target samples as an input to `mutations`, you will get an error `Collection <collection> has no data:mutationstable data!`, even though data objects with type `data:mutationstable` exist in this collection. It would be better if this error messages were more informative. Example of an error: 

```
vt = VariantTables(c, mutations=["TP53: Ser"])
vt.variants
ValueError: Collection <collection> has no data:mutationstable data!
```

src/resdk/tables/variant.py
```diff
@@ -0,0 +1,424 @@
...
+
+        def chrom_pos(column_name):
+            """Return only chromosome and position from index."""
+            chrom, pos = column_name.split("_")[:2]
+            return f"{chrom}_{pos}"
+
+        to_discard = list()
+
+        for _, group in groupby(df.columns, key=chrom_pos):
+            # Sorting moves fake variant to the first position
```
OtonicarJan, line 396: There is an issue if you first call `VariantTables` with `discard_fakes` set to default (False) and then try to call `VariantTables` again but with `discard_fakes` set to True. "fakes" will not be removed in the second call, you first have to clear the cache. The default value for `discard_fakes` could be set to True.

OtonicarJan (approved):
Amazing effort! Left two suggestions for improvement.

gregorjerse (approved):

PR #294, opened 2022-09-13
Support setting descriptor on Relation resource

JureZmrzlikar:


jkokosar (approved):
The feature works as expected!

src/resdk/resources/relation.py
```diff
@@ -77,9 +88,20 @@ def collection(self, payload):
...
+
+    @descriptor_schema.setter
+    def descriptor_schema(self, payload):
+        """Set descriptor schema."""
+        self._resource_setter(payload, DescriptorSchema, "_descriptor_schema")
+
     def update(self):
         """Clear cache and update resource fields from the server."""
         self._samples = None
+        self._descriptor_schema = None
```
gregorjerse, line 104: Why only clear `_descriptor_schema` here and no `_descriptor`?

gregorjerse (approved):

src/resdk/resources/relation.py
JureZmrzlikar: Field `descriptor_schema` is a property, and the contents of it are stored in `_descriptor_schema` attribute. Field `descriptor` is a totally "normal" attribute, thus handling / caching of these values are handled differently. See for example, how we handle descriptor schema in `Data` or `Collection` Resource - it is the same as introduced in this commit.

PR #296, opened 2022-10-04
Use direct path to objects uploaded to S3 bucket

gregorjerse:
instead of generating a presigned url


JureZmrzlikar (approved):
Looks good and works for me. Thanks!

PR #298, opened 2022-10-05
Use posix style paths when uploading to S3

gregorjerse:


JureZmrzlikar (commented):
LGTM, but i do not posses Windows system, so I will wait for a user who does to approve.

jkokosar (approved):
The fix was tested on Win10

JureZmrzlikar (approved):

PR #300, opened 2022-10-28
Rework VariantTables

JureZmrzlikar:


OtonicarJan (commented):
There are currently some issues with the number of reported variants by default - the number of mutations for a gene is lower than if you only call `VariantTables` with `mutations=[gene]`. I also get `TimeoutError` if I call `VariantTables` with `mutations=[genes from gene set]`.

JureZmrzlikar:
Hi @OtonicarJan 

I've made the modifications to this PR hat we discussed about. Can you please test the code on relevant datasets?
Also, please check if this is still an issue and if so, please provide steps to reproduce.
> There are currently some issues with the number of reported variants by default - the number of mutations for a gene is lower than if you only call VariantTables with mutations=[gene]. 

Thanks!

src/resdk/tables/variant.py
```diff
@@ -99,70 +99,118 @@ def __init__(
...
+    def depth_g(self) -> pd.DataFrame:
+        """Return depth table for guanine as a pandas DataFrame object."""
+        return self._load_fetch(self.DEPTH_G)
+
+    @property
+    @lru_cache()
+    def depth_t(self) -> pd.DataFrame:
+        """Return depth table for thymine as a pandas DataFrame object."""
+        return self._load_fetch(self.DEPTH_T)
+
```
OtonicarJan, line 170: Add additional function to compute dataframe based on column `FILTER` from input dataframes. The FILTER column tells us if the variant passed variant filtration, so something like `tables.qc` could be used to return the dataframe. Values in FILTER column are strings.

src/resdk/tables/variant.py
```diff
@@ -279,20 +280,14 @@ def construct_index(self, row) -> str:
...
-        aa_change = row.get("HGVS.p", "?")
-        aa_change = self.strip_amino_acid(aa_change)
-        if aa_change:
-            index = f"{index}_{aa_change}"
-
         return index
 
     @staticmethod
-    def encode_mutation(row) -> int:
+    def _encode_mutation(row) -> int:
```
OtonicarJan, line 290: After the latest change to process `mutations-table`, variants are reported only if they have been detected. This means that you cannot get wild type variant in input samples - you will never have a zero in `tables.variants`. I suggest that we remove encoding for wild type.

src/resdk/tables/variant.py
```diff
@@ -379,15 +359,20 @@ def _parse_file(self, file_obj, sample_id, data_type) -> pd.Series:
...
         s.name = sample_id
+
+        # It si common for mutations_table.tsv to contains duplicate variants
+        # For now, eep the first one and drop the rest of them.
+        s = s[~s.index.duplicated(keep="first")]
+
         return s
 
-    def postprocess_variants_table(self, df, data_type):
+    def _postprocess_variants_table(self, df, data_type):
```
OtonicarJan, line 369: Because of the change in `mutations-table` mentioned in the previous comment, I think we can remove this function. There will not be any positions reported for which we know that do not contain variant by default. Furthermore, if two variants are present in the collection for the same position but different nucleotide change, we would lose one of them if we discarded "fake variant".

docs/resdk-tables.rst
```diff
@@ -153,17 +153,17 @@ access of variant data present in Data of type ``data:mutationstable``::
...
-sample_id  chr1_123_C>T_Gly11Asp  chr1_126_T>C_Asp12Gly
-=========  =====================  =====================
-101        2                      0
-102        0                      1
-=========  =====================  =====================
+=========  ============  ============
+sample_id  chr1_123_C>T  chr1_126_T>C
+=========  ============  ============
+101        2             0
+102        0             1
```
OtonicarJan, line 160: Update according to comments below - 0 is not possible anymore.

docs/resdk-tables.rst
```diff
@@ -179,12 +179,12 @@ can also be ``NaN`` (unknown). Other common case is just the info that
...
 also. Given the above, a more realistic example of output is:
 
-=========  =====================  ===================== ========
-sample_id  chr1_123_C>T_Gly11Asp  chr1_126_T>C_Asp12Gly chr1_127
-=========  =====================  ===================== ========
-101        2                      NaN                   0
-102        0                      1                     NaN
-=========  =====================  ===================== ========
+=========  ============  ============ ========
+sample_id  chr1_123_C>T  chr1_126_T>C chr1_127
```
OtonicarJan, line 183: Variants like `chr1_127` (only position present without nucleotide change) are not possible anymore since the latest change to `mutations-table`. This kind of variants were reported only in cases where there was no variant present in the sample but was present in the reference file.

src/resdk/tables/variant.py
```diff
@@ -205,71 +256,21 @@ def _data(self) -> List[Data]:
...
 
         Index should have the form:
-        <chr>_<position>_<snp_change>_<aminoacid_pos_change>
-        E.g. chr2_1234567_C>T_Gly12Asp
+        <chr>_<position>_<snp_change>
+        E.g. chr2_1234567_C>T
 
-        In cases where nucleotide change or amino-acid change is
-        missing, this info is left out from the index.
+        In cases where nucleotide change is missing, this info is left out from
```
OtonicarJan, line 272: This can never happen anymore, since the variant is only reported if it was detected.

src/resdk/tables/variant.py
```diff
@@ -99,70 +99,118 @@ def __init__(
...
+        # Geneset can be set only once, prevent modifications
+        if self._geneset is not None:
+            raise ValueError("It is not allowed to change geneset value.")
+
+        if value is None:
+            return
+
+        # If id / slug of a geneset is given, get it from the Resolwe server
+        if isinstance(value, (int, str)):
+            gs = self.resolwe.data.get(value)
```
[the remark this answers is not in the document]
JureZmrzlikar: Both wor, but geneset is more specific: geneset is only a special type of Data. Will fix it, thanks.

src/resdk/tables/variant.py
```diff
@@ -99,70 +99,118 @@ def __init__(
...
+        return self._load_fetch(self.DEPTH_T)
+
+    def _check_heterogeneous_mutations(self):
+        """Check there are not heterogeneous mutations / genesets.
+
+        Genes for which mutations are computed are given either with mutations
+        (list of genes) or geneset (geneset ID) input. Ensure all the data has
+        the same value of this.
+        """
+        mutations = {str(d.input.get("mutations", "")) for d in self._data}
```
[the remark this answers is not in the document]
JureZmrzlikar: Could this be solved with:
`mutations = {str(d.input.get("mutations", [])) for d in self._data}`

src/resdk/tables/variant.py
JureZmrzlikar: Cool!

src/resdk/tables/variant.py
JureZmrzlikar: <3

src/resdk/tables/variant.py
```diff
@@ -79,90 +80,185 @@ class VariantTables(BaseTables):
...
+
+        Values can be:
+
+            - PASS - Variant has passed filters:
+            - DP : Insufficient read depth
+            - QD: insufficient quality normalized by depth
+            - FS: insufficient phred-scaled p-value using Fisher's exact
+                test to detect strand bias
+            - SnpCluster: Variant is part of a cluster
+
```
OtonicarJan, line 208: Maybe you can add specific filters used: FS > 30.0, QD < 2.0, DP < 10.0. So for example, if a variant has DP of 8, GATK will replace PASS with DP in FILTER column.

OtonicarJan (approved):
Looks and works great!

docs/CHANGELOG.rst
```diff
@@ -26,6 +26,14 @@ Changed
   - ``Sample.get_cuffquant``
   - ``Sample.get_expression``
 
+- **BACKWARD INCOMPATIBLE:** Rework ``VariantTables``:
+
+  - Index in VariantTables.variants is simplified and does not include
+    ammino-acid change anymore.
+  - Argument ``mutations`` in ``VariantTables`` constructor is renamed to
+    ``geneset``. Besides holding a list fo genes, this can also be a valid ID /
```
gregorjerse, line 34: Should ID be capitalized? Maybe id/slug is more readable.

docs/resdk-tables.rst
```diff
@@ -153,49 +153,23 @@ access of variant data present in Data of type ``data:mutationstable``::
...
-=========  =====================  =====================
+=========  ============  ============
+sample_id  chr1_123_C>T  chr1_126_T>C
+=========  ============  ============
+101        2             NaN
+102        0             2
+=========  ============  ============
 
 
 In rows, there are sample ID's. In columns there are variants where each
```
gregorjerse, line 164: I see... ID is always capitalized?

src/resdk/tables/variant.py
```diff
@@ -79,90 +80,187 @@ class VariantTables(BaseTables):
...
-            if not isinstance(mutations, list):
-                raise ValueError("Mutations must be a list of strings")
+    @property
+    def geneset(self):
+        """Get geneset."""
+        return self._geneset
+
+    @geneset.setter
+    def geneset(self, value):
+        """Set geneset."""
```
gregorjerse, line 119: Add a type hint for value (int, str or Geneset if I am not mistaken).
Add some comment about the value attribute: it can either be id, slug or a Geneset instance.

gregorjerse: Also add comment that the geneset can be set only once and when ValueError is raised.

src/resdk/tables/variant.py
```diff
@@ -79,90 +80,187 @@ class VariantTables(BaseTables):
...
 
-    def check_heterogeneous_mutations(self):
-        """Ensure variants are selected for the same mutations in all samples."""
-        mutations = {str(d.input["mutations"]) for d in self._data}
+    @property
+    @lru_cache()
+    def depth_g(self) -> pd.DataFrame:
+        """Get depth table for guanine."""
+        return self._load_fetch(self.DEPTH_G)
+
```
gregorjerse, line 187: Just curious: is it easier to user five methods (depth[acgt]?) or just one (depth) with optional argument specifying the amino acid?

gregorjerse: I see that the methods were already  in use before, so please ignore the previous question.

gregorjerse (approved):

docs/CHANGELOG.rst
JureZmrzlikar: We use capitalized version quite commonly.

docs/resdk-tables.rst
JureZmrzlikar: :+1:

src/resdk/tables/variant.py
JureZmrzlikar: Done.

PR #301, opened 2022-11-28
Add build property to RNATables

JureZmrzlikar:


LeaVohar:
I tested several collections and it works well.

LeaVohar (approved):

src/resdk/tables/rna.py
```diff
@@ -366,6 +366,20 @@ def readable_columns(self) -> Dict[str, str]:
 
         return self._mapping(self.gene_ids, source, species)
 
+    @property
+    @lru_cache()
+    def build(self) -> str:
+        """Get build."""
+        builds = list(set(d.output.get("build") for d in self._data))
```
gregorjerse, line 373: I do not fully understand what you are trying to achieve here.
1. The `d.output.get("build")` will always be at least None, so `len(builds)` will always be at least 1.
2. If you only want to check if there is precisely one build, then the last cast to list is probably redundant.

gregorjerse (approved):

src/resdk/tables/rna.py
JureZmrzlikar: > The d.output.get("build") will always be at least None, so len(builds) will always be at least 1.

Not if `self._data` is empty list. 

> last cast to list is probably redundant

The cast is done here to avoid casting to list in the final line of the function. Matter of preference.

PR #302, opened 2022-11-30
Optionally show progress bar in resdk.Query.iterate()

JureZmrzlikar:


mzganec (approved):
This implementation is correct. If we use `mininterval=0` then the ETA on the right side of the progress bar is incorrect because it estimates the remaining time based on the difference from the previous `tqdm` update. The pull request is good to merge.

PR #309, opened 2023-04-05
Fix/e2e tests

gregorjerse:


src/resdk/resources/background_task.py
```diff
@@ -0,0 +1,73 @@
...
+    :param resolwe: Resolwe instance
+    :type resolwe: Resolwe object
+    :param model_data: Resource model data
+
+    """
+
+    endpoint = "task"
+
+    READ_ONLY_FIELDS = BaseResource.READ_ONLY_FIELDS + (
+        "pk",
```
JureZmrzlikar, line 22: Didd you add this by mistake or is it not documented below?

src/resdk/resources/background_task.py
```diff
@@ -0,0 +1,73 @@
...
+        self.finished = None
+        #: status - Possible values:
+        #: WA (waiting)
+        #: PR (processing)
+        #: OK (done)
+        #: ER (error)
+        self.status = None
+        #: description
+        self.description = None
+        #: output
```
JureZmrzlikar, line 47: Would be nice to know / document what the output is. A string, number, json with complex structure, etc...?

JureZmrzlikar (approved):
LGTM, just two minor comments to address.
Thanks for your effort!

src/resdk/resources/background_task.py
gregorjerse: Actually a bit of both. The serializer in the Resolwe class returned `pk` instead of `id`, which is now fixed.

src/resdk/resources/background_task.py
gregorjerse: I have added a short comment. It is JSON but the actual value depends on the background task. Currently, only the duplicate task is supported and for it, the value can be:
- a list of ids of duplicated objects on success
- a string with error details on failure

Some other background tasks (in the future) may use output differently.

src/resdk/resources/background_task.py
```diff
@@ -0,0 +1,72 @@
...
+        #: ER (error)
+        self.status = None
+        #: description
+        self.description = None
+        #: output - JSON field, the actual value depends on the background task
+        self.output = None
+
+        super().__init__(resolwe, **model_data)
+
+    def result(self, timeout=10, refresh_interval=1.0, final_statuses=["OK"]):
```
dblenkus, line 51: I suggest renaming the method to `wait` and removing all the validation. It will be much easier to integrate websockets once they are supported in the ReSDK. I would also remove the `refresh_interval` as it won't make sense with websockets, and we don't need it now.

src/resdk/resources/data.py
```diff
@@ -156,10 +156,14 @@ def descriptor_schema(self, payload):
     def sample(self):
         """Get sample."""
         if self._sample is None and self._original_values.get("entity", None):
+            # The collection data is only serialized on the top level. Replace the
+            # data inside 'entity' with the actual collection data.
+            self._original_values["entity"]["collection"] = self._original_values.get(
```
dblenkus, line 161: Do you know if we need this? If not, I would prefer to keep the "original data" actually original.

gregorjerse: OK, I will construct another dict and avoid modifying the original values.

src/resdk/resources/background_task.py
gregorjerse: OK, done.

dblenkus (approved):

PR #313, opened 2023-06-23
Fix RNATables in case of very large collecitons

JureZmrzlikar:


LeaVohar (approved):
Thanks Jure, it works. It takes about 25 minutes to get .rc and 35 minutes to get expressions. I tried running the script and in ipython environment and both work fine.

gregorjerse (approved):

PR #320, opened 2023-08-28
Replace username with email where necessary

JureZmrzlikar:


gregorjerse (approved):
I am not sure we need to change the username to email in the code. 

It is just a "simple" change: the username now is email, so it can still be named username in the code.

PR #323, opened 2023-08-30
Show inherited members members in docs/ref

JureZmrzlikar:


klemenziberna (approved):
Looks good to me. I like to use documentation as my primary source of truth (vs. using IDE to explore implemented methods), so this will be of great use to me. Thank you for adding this!

gregorjerse (approved):

PR #324, opened 2023-08-31
Warn user of missing gene ID / symbol mappings

JureZmrzlikar:


src/resdk/tables/rna.py
```diff
@@ -414,7 +414,16 @@ def readable_columns(self) -> Dict[str, str]:
...
-        return self._mapping(self.gene_ids, source, species)
+        mapping = self._mapping(self.gene_ids, source, species)
+        if len(mapping) < len(self.gene_ids):
+            missing_count = len(self.gene_ids) - len(mapping)
+            warnings.warn(
+                f"Symbols for {missing_count} gene IDs were not found. "
+                "Missing symbols will be set to empty string.",
+                UserWarning,
+            )
+            mapping = {gene_id: mapping.get(gene_id, "") for gene_id in self.gene_ids}
```
klemenziberna, line 425: This is more for discussion. How do we handle missing information in our repositories?

While empty string ("") is definitely one of the options, have we considered other alternatives:
- e.g. using None (SQL-like behaviour, supported by pandas),
- using pd.NA (supported by pandas),
- using more explicit string, e.g."unknown" or even "NaN"?

I personally would prefer one of the pandas-supported options, so that works well with common approaches to check and handle NaNs (value_counts, isna, fillna, etc.), but either is fine as long as it is clearly documented and (ideally) consistent across our code base.

klemenziberna (approved):

src/resdk/tables/rna.py
JureZmrzlikar: Good point, thanks for raising it. 

I opted for empty string to keep the type of all columns string. It's difficult to say that we have a very consistent way of  using one of the above NA options, but IMO the most dominant is the usage of `np.nan`, which is also nicely supported by `pandas`. I suggest we go with it.

gregorjerse (approved):

PR #325, opened 2023-10-05
Fix VariantTables in case genes are missing from KB

JureZmrzlikar:


src/resdk/tables/variant.py
```diff
@@ -254,7 +254,18 @@ def _get_obj_geneset(self, obj):
             if gs.output["source"] != "UCSC":
                 qs = self.resolwe.feature.filter(feature_id__in=list(obj_geneset))
                 id_2_name = {obj.feature_id: obj.name for obj in qs}
-                obj_geneset = set([id_2_name[gene] for gene in obj_geneset])
+                obj_geneset = set(
```
gregorjerse, line 257: Do we have a lot of names in obj_geneset? Here we iterate over the list twice, but we could do everything in a single pass.

gregorjerse: Another remark: creating a list inside a set is not necessary.

src/resdk/tables/variant.py
```diff
@@ -254,7 +254,18 @@ def _get_obj_geneset(self, obj):
...
+                    [id_2_name[gene] for gene in obj_geneset if gene in id_2_name]
+                )
+                # Sometimes, geneset defined in obj.input[geneset/mutations] are
+                # missing in KnowledgeBase. This can happen due to KB updates.
+                genes_with_missing_symbols = [
+                    gene for gene in obj_geneset if gene not in id_2_name
+                ]
```
gregorjerse, line 264: This will not work as expected: the `obj_geneset` above contains only genes in `id_2_name` so the condition in the list comprehension expression will never evaluate to True.

gregorjerse (changes requested):

gregorjerse (approved):

PR #326, opened 2023-10-09
Explicitly set dtypes of all columns in RNATabnles.qc

JureZmrzlikar:


klemenziberna (approved):
This solves my issue with saving collections to AnnData that contain QC metadata.

PR #327, opened 2023-10-19
Feature/streamlined annotations

gregorjerse:


src/resdk/query.py
```diff
@@ -92,6 +98,10 @@ def __init__(self, resolwe, resource, slug_field="slug"):
 
         self.logger = logging.getLogger(__name__)
 
+    def _non_string_iterable(self, item) -> bool:
+        """Return thur when item is iterable but not string."""
```
JureZmrzlikar, line 102: Spelling: "thur" should be changed to True, i guess?

Also, could you explain a bit what this function is necessary?

src/resdk/query.py
```diff
@@ -380,3 +388,52 @@ def iterate(self, chunk_size=100, show_progress=False):
...
+
+    def _fetch(self):
+        """Make request to the server and populate cache.
+
+        Fetch all values and their fields with 2 queries.
+        """
+        # Execute the query in a single request.
+        super()._fetch()
+
+        missing = {
```
JureZmrzlikar, line 416: Cold you explain why this scenario happens? Why in some cases `self._field` is None, while in others it is not None?

src/resdk/resolwe.py
```diff
@@ -166,10 +177,13 @@ def _initialize_queries(self):
         """Initialize ResolweQuery's."""
         for resource, query_name in self.resource_query_mapping.items():
             slug_field = self.slug_field_mapping.get(query_name, "slug")
-            query = ResolweQuery(self, resource, slug_field=slug_field)
+            QueryClass = self.resource_query_class.get(resource, ResolweQuery)
+            query = QueryClass(self, resource, slug_field=slug_field)
             if query_name in self.query_filter_mapping:
                 query = query.filter(**self.query_filter_mapping[query_name])
             setattr(self, query_name, query)
+        # Use custon query to reduce the number of queries.
```
JureZmrzlikar, line 185: Typo (custon).

JureZmrzlikar: Could you please explain, I don;t see any difference between what has been done in the loop above?

src/resdk/resources/__init__.py
```diff
@@ -41,6 +41,14 @@
 .. autoclass:: resdk.resources.DescriptorSchema
    :members:
    :inherited-members:
+   
+.. autoclass:: resdk.resources.AnnotationValue
```
JureZmrzlikar, line 45: consider adding the AnnotationGroup as well, as it is used, e.g. in the `AnnotationField.group`

src/resdk/resources/annotations.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class AnnotationValue(BaseResource):
+    """Resolwe AnnotationValue resource."""
+
+    endpoint = "annotation_value"
+
+    READ_ONLY_FIELDS = BaseResource.READ_ONLY_FIELDS + ("label",)
+
+    UPDATE_PROTECTED_FIELDS = BaseResource.UPDATE_PROTECTED_FIELDS + ("entity", "field")
```
JureZmrzlikar, line 81: The filed `entity` is called sample in resdk. Could you enable that users would be able to relate to the sample as `AnnotationValue.sample`? However, this will only return Sample ID. So maybe it is better to expose this as `AnnotationValue.sample_id`... We can discuss this live.

src/resdk/query.py
gregorjerse: Before there were checks where it was expected for resources to be of instance list (`_dehydrate_resources`, `_add_filter`). I believe I used them with iterable (dictkeys...) or something like that so I changed it to a more general type.

src/resdk/query.py
gregorjerse: This is only optimization: the annotation fields are fetched in a single query when reading corresponding annotation values.

Maybe this is not necessary. Consider the following use case:

```
values = res.annotation_values.filter(entity=xxx)
for value in values:
    print(value.field.name, value.field.group.name)
```

This would query the backend for every field (so it would produce n queries). The "optimized" fetch only produces two queries (one for values and one for fields). One could argue that fields are not always used so the second query is not always necessary.

src/resdk/resolwe.py
gregorjerse: This is an old code introduced before the `resource_query_class`. It is redundant and has to be removed.

src/resdk/resources/__init__.py
gregorjerse: Sure.

src/resdk/query.py
```diff
@@ -380,3 +388,52 @@ def iterate(self, chunk_size=100, show_progress=False):
...
+                missing[field.id]._field = field
+                missing[field.id]._original_values["field"] = field._original_values
+
+    def from_path(self, full_path: str) -> "AnnotationValue":
+        """Get the AnnotationValue from full path.
+
+        :raises LookupError: when field at the specified path does not exist.
+        """
+        group_name, field_name = full_path.split(".", maxsplit=1)
+        return self.get(field__name=field_name, field__group__name=group_name)
```
JureZmrzlikar, line 432: This will probably return all AnnotationValues with given field. So it is probably `self.filter(...)` instead of `self.get(...)`

src/resdk/resources/annotations.py
```diff
@@ -0,0 +1,136 @@
...
+
+class AnnotationValue(BaseResource):
+    """Resolwe AnnotationValue resource."""
+
+    endpoint = "annotation_value"
+
+    READ_ONLY_FIELDS = BaseResource.READ_ONLY_FIELDS + (
+        "field",
+        "entity",
+        "value",
```
JureZmrzlikar: Fields `field` and `value` are among other fields already.... Also, why is `entity` here?

src/resdk/resources/annotations.py
```diff
@@ -0,0 +1,136 @@
...
+        """Get annotation field."""
+        if self._field is None:
+            assert (
+                self.field_id is not None
+            ), "AnnotationField must be set before it can be used."
+            self._field = AnnotationField(self.resolwe, id=self.field_id)
+        return self._field
+
+    @field.setter
+    def field(self, payload):
```
JureZmrzlikar: Consider raising error in case `_field` is not None. This would prevent users from believing they updated the field immediately. I realize that this would raise error during `save()`

src/resdk/query.py
JureZmrzlikar: I see, the backend refuses to provide result unless one entity filter is provided. However, my comment is still relevant, consider this filter:
`res.annotation_value.filter(entity__collection=<collection>).from_path("general.species")`
This will result in many results and `AnnotationValueQuery.get` fails since it got multiple results.

Please add a comment in docstring that indicates that user needs to provide a sample (entity) filter.

JureZmrzlikar (commented):
Very close! :-)
Just a few comments.

JureZmrzlikar (approved):
Looks good, thanks for you effort!

PR #333, opened 2023-11-14
Add bulk annotation helper on sample

gregorjerse:


src/resdk/query.py
```diff
@@ -222,6 +222,9 @@ def _fetch(self):
         if isinstance(items, dict) and "results" in items:
             self._count = items["count"]
             items = items["results"]
+        # Store count when list of objects is received.
+        if isinstance(items, list):
+            self._count = len(items)
```
dblenkus, line 227: 👌🏼

mzganec:
Works well.

Feature request:
- Consider implementing also `Sample.get_annotations`, which would return all annotated fields in a dict. This type of output is more useful to bioinformaticians than the list of key-value pairs that we get from `Sample.annotations.all()`.

Bugs:
- Both `Sample.annotations.count()` and `Sample.annotations.all().count()` return a wrong number of items. I annotated four fields and `count()` returned one.
- `s.set_annotation` (note the singular form) fails if the value is set to `None`, i.e., deleting a field is not supported.

gregorjerse:
1. Helper get_annotations is implemented.
2. Bugs should be fixed now.

mzganec (approved):
Well done! Works as expected.

I noticed that `set_annotation` (note the singular form) requires explicit confirmation when deleting a field. That's fine, but note that bioinformaticians typically annotate all samples in a collection, so this may require them to press "y" several hundred times. A `force` argument would be nice to control the confirmation prompt. This issue doesn't need to be addressed immediately because `set_annotations` doesn't prompt for confirmation, so we can use that instead.

PR #345, opened 2024-02-01
Invalidate resdk.tables cache on AnnotationValue updates

JureZmrzlikar:


marcellevstek (approved):
QA done on the PR. Works as expected.

src/resdk/resources/annotations.py
```diff
@@ -92,7 +94,7 @@ class AnnotationValue(BaseResource):
 
     endpoint = "annotation_value"
 
-    READ_ONLY_FIELDS = BaseResource.READ_ONLY_FIELDS + ("label", "modified")
+    READ_ONLY_FIELDS = BaseResource.READ_ONLY_FIELDS + ("label",)
```
gregorjerse, line 97: Why was the field `modified` removed?

src/resdk/resources/annotations.py
```diff
@@ -115,6 +117,12 @@ def __init__(self, resolwe: "Resolwe", **model_data):
         self._sample: Optional[Sample] = None
         super().__init__(resolwe, **model_data)
 
+    @property
+    @assert_object_exists
+    def modified(self):
+        """Modification time."""
+        return parse_resolwe_datetime(self._original_values["modified"])
```
gregorjerse, line 124: Just a remark: I believe Resolwe uses ISO 8601 datetime. So you could just use `datetime.fromisoformat(date_string)` to convert from it to the `datetime` object.

gregorjerse (approved):

src/resdk/resources/annotations.py
JureZmrzlikar: OK, will check this solve in one of the subsequent PRs.

PR #346, opened 2024-02-02
Improve download

JureZmrzlikar:


src/resdk/resolwe.py
```diff
@@ -531,6 +534,18 @@ def _download_files(
                             file_handle.write(chunk)
                             progress_bar.update(len(chunk))
 
+                # Verify md5 checksum:
+                if file_name.endswith(".html"):
+                    # Due to backend processing, html file fields have
+                    # checksums that are difficult to reproduce here.
+                    return
+                expected_md5 = checksums[file_directory][file_name]
+                computed_md5 = md5(os.path.join(download_dir, file_path, file_name))
```
gregorjerse, line 543: It would be more efficient to integrate md5 checksum computation in the download process and compute it for each downloaded chunk.

However, computing it after the file is written to disk also captures disk write errors (or, in case of BCM, NFS errors).

gregorjerse (approved):

PR #352, opened 2024-03-18
Add variant related models

gregorjerse:
REID-679

### DIFF
--- a/src/resdk/resources/variants.py
+++ b/src/resdk/resources/variants.py
@@ -1,0 +1,274 @@
+"""Variant resources."""
+
+from datetime import datetime
+from typing import Any
+
+from .base import BaseResource
+from .serializer import ResourceNestedDictSerializer
+
+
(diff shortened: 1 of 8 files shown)


src/resdk/resources/variants.py
```diff
@@ -0,0 +1,73 @@
...
+        "quality",
+        "depth",
+        "filter",
+        "genotype",
+        "data",
+    )
+
+    def __repr__(self) -> str:
+        """Return string representation."""
+        return f"VariantCall <path: {self.field.group.name}.{self.field.name}, value: '{self.value}'>"
```
JureZmrzlikar, line 73: Are you sure this is OK? I do not see any fields with name `field` or `value` in this class?

gregorjerse: No, this is a mistake. Fixed.

gregorjerse:
Rewrite in #381

PR #353, opened 2024-04-08
Replace int64 with nullable Int64 in RNATables

AnzeLovse:
Calling RNATables.qc would fail with `Cannot convert non-finite values (NA or inf) to integer: Error while type casting for column 'star_assigned_reads'` in cases where MultiQC data is not available for a subset of samples. 

All other integer columns use `Int64` dtype which supports missing values (https://pandas.pydata.org/docs/user_guide/integer_na.html).

### DIFF
--- a/docs/CHANGELOG.rst
+++ b/docs/CHANGELOG.rst
@@ -14,6 +14,12 @@ Added
 - Add ``restart`` method to the ``Data`` resource
 
 
+Fixed
+-----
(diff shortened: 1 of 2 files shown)


AnzeLovse:
@JureZmrzlikar would you please review this.

JureZmrzlikar (approved):
LGTM, thanks for fixing this!

PR #360, opened 2024-07-11
Use NULL_EXECUTOR in e2e tests

gregorjerse:


JureZmrzlikar (approved):
LGTM.

.github/workflows/e2e.yaml
```diff
@@ -105,8 +105,28 @@ jobs:
 
       - name: Try the cluster
         run: |
-          kubectl wait --for=condition=available --timeout=1200s -n default deployment/postgres
           # Until postgres is fully restored, the services will fail
+          kubectl wait --for=condition=available --timeout=1200s -n default deployment/postgres
+          # Patch the channels manager to use the null executor
+          kubectl patch deployment channels-manager -n default --patch '{
```
acopar, line 111: Could we read the patch from a json file instead of inline?

PR #361, opened 2024-09-30
Add functionality for setting custom names

marcellevstek:
of downloaded files

### DIFF
--- a/src/resdk/resolwe.py
+++ b/src/resdk/resolwe.py
@@ -543,7 +543,7 @@ class Resolwe:
                 computed_md5 = md5(os.path.join(download_dir, file_path, file_name))
                 if expected_md5 != computed_md5:
                     raise ValueError(
-                        f"Checksum of downloaded file {file_name} does not match the expected value."
+                        f"Checksum ({computed_md5}) of downloaded file {file_name} does not match the expected value of {expected_md5}."
                     )
 
     def data_usage(self, **query_params):
--- a/src/resdk/resolwe.py
+++ b/src/resdk/resolwe.py
@@ -462,7 +462,10 @@ class Resolwe:
         return Data(resolwe=self, **model_data)
 
     def _download_files(
-        self, files: List[Union[str, Path]], download_dir=None, show_progress=True
+        self,
+        files: List[Union[str, Path]],
+        download_dir: Union[str, None] = None,
+        show_progress: bool = True,
     ):
         """Download files.
 
@@ -470,10 +473,8 @@ class Resolwe:
         directory (defaults to the current working directory).
 
         :param files: files to download
-        :type files: list of file URI
         :param download_dir: download directory
-        :type download_dir: string
-        :rtype: None
+            If not specified, the current working directory is used.
 
         """
         if not download_dir:
@@ -486,7 +487,6 @@ class Resolwe:
 
         if not files:
             self.logger.info("No files to download.")
-
         else:
             self.logger.info("Downloading files to %s:", download_dir)
             # Store the sizes of files in the given directory.
--- a/src/resdk/resources/data.py
+++ b/src/resdk/resources/data.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+import os
 from typing import Optional
 from urllib.parse import urljoin
 
@@ -324,19 +325,16 @@ class Data(BaseResolweResource):
 
         return file_list
 
-    def download(self, file_name=None, field_name=None, download_dir=None):
+    def download(
+        self,
+        file_name: Optional[str] = None,
+        field_name: Optional[str] = None,
+        download_dir: Optional[str] = None,
+    ):
         """Download Data object's files and directories.
 
         Download files and directories from the Resolwe server to the
         download directory (defaults to the current working directory).
-
-        :param file_name: name of file or directory
-        :type file_name: string
-        :param field_name: file or directory field name
-        :type field_name: string
-        :param download_dir: download path
-        :type download_dir: string
-        :rtype: None
 
         Data objects can contain multiple files and directories. All are
         downloaded by default, but may be filtered by name or output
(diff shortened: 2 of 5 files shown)


src/resdk/resources/data.py
```diff
@@ -353,7 +357,21 @@ def download(self, file_name=None, field_name=None, download_dir=None):
...
         ]
-        self.resolwe._download_files(files, download_dir)
+
+        # Only applies if downloading a single file
+        custom_file_names = None
+        if custom_file_name:
+            if file_name or field_name:
+                custom_file_names = [custom_file_name] * len(files)
+            else:
+                logging.warning(
```
JureZmrzlikar, line 367: I'd raise error, instead of logging a warning. In the case you provide custom file name, but do not specify the file_name / field_name (and there are multiple files on Data), errors will be raised anyway, but might not be as meaningful.

tests/unit/test_data.py
```diff
@@ -178,13 +178,17 @@ def test_download_ok(self, data_mock):
...
         )
 
         data_mock.reset_mock()
         Data.download(data_mock, download_dir="/some/path/")
         data_mock.resolwe._download_files.assert_called_once_with(
-            ["123/file1.txt", "123/file2.fq.gz"], "/some/path/"
+            files=["123/file1.txt", "123/file2.fq.gz"],
+            download_dir="/some/path/",
+            custom_file_names=None,
         )
```
JureZmrzlikar, line 193: Would be nice if you add a test as well. If you need help, don't hesitate to ask.

JureZmrzlikar (commented):
Great work, PR looks ready - technical wise. However, I have second thoughts regaridn the arguments. One way to provide custom name is the way you did:
```
Data.download(
    field_name="bam",
    download_dir="/my/dir/",
    custom_file_name="foo.txt"
)
```
The other option would be to make a breaking change:
```
Data.download(
    field_name="bam",
    target="/my/dir/foo.txt",  # If one just wants to specify dir, target= "/my/dir/"
)
```
The naming `target`, resembles `Path.rename`, we could also use `dst` as in `os.rename`. What do you think? Do you have any opinion @gregorjerse ?

marcellevstek:
@JureZmrzlikar I agree with you about making the breaking change. I didn't want to incorporate it in the current implementation, but it makes complete sense not to use two distinct arguments for constructing the path.

@gregorjerse please review.

src/resdk/resolwe.py
```diff
@@ -462,7 +462,11 @@ def get_or_run(self, slug=None, input={}):
         return Data(resolwe=self, **model_data)
 
     def _download_files(
-        self, files: List[Union[str, Path]], download_dir=None, show_progress=True
+        self,
+        files: List[Union[str, Path]],
+        download_dir=None,
+        show_progress=True,
+        custom_file_names: Union[List[str], None] = None,
```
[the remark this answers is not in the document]
marcellevstek: It would be better to make a mapping of field names to custom file names. If this would be suitable, I can make the change. The problem is that you do not know the file names in advance and this would especially prove to be difficult when downloading data for multiple samples.

src/resdk/resources/collection.py
```diff
@@ -140,7 +140,7 @@ def download(self, file_name=None, field_name=None, download_dir=None):
             data_files = data.files(file_name, field_name)
             files.extend("{}/{}".format(data.id, file_name) for file_name in data_files)
 
-        self.resolwe._download_files(files, download_dir)
+        self.resolwe._download_files(files=files, download_dir=download_dir)
```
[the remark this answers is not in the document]
marcellevstek: Better practice to explicitly name the arguments when calling the function. Not really needed IMO.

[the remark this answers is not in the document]
gregorjerse: Please remove it as it has nothing to do with this PR.

marcellevstek:
@gregorjerse I have added a new function for downloading and renaming a single file. Can you please guide me how to create a test, as I have problem mocking the download function.

src/resdk/resources/data.py
```diff
@@ -353,7 +357,49 @@ def download(self, file_name=None, field_name=None, download_dir=None):
...
+
+        new_file_path = os.path.join(download_dir, custom_file_name)
+
+        if os.path.exists(new_file_path):
+            logging.warning(
+                f"File with path '{new_file_path}' already exists. Skipping download."
+            )
+            return
+
+        file_names = self.download(
```
marcellevstek, line 386: Could be called by constructing kwargs. @gregorjerse if current implementation is not okay, I can change it.

src/resdk/resources/data.py
```diff
@@ -354,20 +357,48 @@ def download(
...
+
+        return file_names
+
+    def download_and_rename(
+        self,
+        custom_file_name: str,
+        field_name: Optional[str] = None,
+        file_name: Optional[str] = None,
+        download_dir: Optional[str] = None,
+    ):
```
gregorjerse, line 372: Just curious: is the use case when you want to download just a single filename really used a lot?

src/resdk/resources/data.py
```diff
@@ -354,20 +357,48 @@ def download(
...
+            return
+
+        file_names = self.download(
+            file_name=file_name,
+            field_name=field_name,
+            download_dir=download_dir,
+        )
+        if len(file_names) != 1:
+            raise ValueError(
+                f"Expected one file to be downloaded, but got {len(file_names)}"
```
gregorjerse, line 393: Here the files have already been downloaded and are left in the download folder. What to do with them?

src/resdk/resources/data.py
```diff
@@ -354,20 +357,48 @@ def download(
...
+            )
+        og_file_name = file_names[0]
```
gregorjerse, line 395: What does `og_file_name` stand for?

src/resdk/resources/data.py
marcellevstek: Yes, I almost exclusively download just one file from a data object.

src/resdk/resources/data.py
```diff
@@ -354,20 +357,48 @@ def download(
...
+    def download_and_rename(
+        self,
+        custom_file_name: str,
+        field_name: Optional[str] = None,
+        file_name: Optional[str] = None,
+        download_dir: Optional[str] = None,
+    ):
+        """Download and rename a single file from data object."""
+
+        if download_dir is None:
```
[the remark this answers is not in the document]
marcellevstek: Would be better, as I am more familiar with it

src/resdk/resources/data.py
marcellevstek: Original file name that is present on the platform.

gregorjerse (approved):

PR #363, opened 2024-10-18
Add a method that outputs server module versions

marcellevstek:


src/resdk/resolwe.py
```diff
@@ -200,6 +200,18 @@ def version_check(self):
                 "Warning: unable to read the minimal supported version from the server."
             )
 
+    def version_output(self):
```
gregorjerse, line 203: Please annotate the method with the return type.

src/resdk/resolwe.py
```diff
@@ -200,6 +200,18 @@ def version_check(self):
...
+        """Output the version of the server modules."""
+        url = urljoin(self.url, "/about/versions")
+        try:
+            response = requests.get(url)
+            server_version = response.json()
+            return server_version
+        except Exception:
+            self.logger.warning(
```
gregorjerse, line 211: This hides the original exception, which may contain important information about the reason the request failed.

I suggest you replace the `self.logger.warning` with `self.logger.exception` and re-raise the exception (or create a new one and raise that one). For instance, create a ResolweServerError with the description bellow joined with the information from the original exception.

src/resdk/resolwe.py
```diff
@@ -200,6 +200,18 @@ def version_check(self):
                 "Warning: unable to read the minimal supported version from the server."
             )
 
+    def version_output(self):
+        """Output the version of the server modules."""
+        url = urljoin(self.url, "/about/versions")
```
gregorjerse, line 205: Please add `/about/versions` as constant. See `MIMIMAL_SUPPORTED_VERSION_POSTFIX` as example (although its naming  is terrible).

gregorjerse (changes requested):

src/resdk/resolwe.py
```diff
@@ -200,17 +201,17 @@ def version_check(self):
...
-            return server_version
-        except Exception:
-            self.logger.warning(
-                "Warning: unable to read the server version from the server."
+        except requests.exceptions.RequestException:
+            raise ResolweServerError(
+                "Unable to read the server version from the server."
             )
+        server_version = response.json()
+        return server_version
```
gregorjerse, line 214: Maybe just `return response.json()` ?

gregorjerse (approved):

PR #364, opened 2024-11-04
Add support for predictions app

gregorjerse:


src/resdk/query.py
```diff
@@ -423,3 +422,40 @@ def _fetch(self):
...
+
+        missing = collections.defaultdict(list)
+        for value in self._cache:
+            if value._field is None:
+                missing[value.field_id].append(value)
+
+        if missing:
+            # Get corresponding annotation field details in a single query and attach it to
+            # the values.
+            for field in self.resolwe.prediction_field.filter(id__in=missing.keys()):
```
dblenkus, line 458: Use `.iterate()` to make it more robust.

src/resdk/resolwe.py
```diff
@@ -106,26 +113,30 @@ class Resolwe:
...
     # Map resource class to ResolweQuery name
     resource_query_mapping = {
         AnnotationField: "annotation_field",
         AnnotationValue: "annotation_value",
-        Data: "data",
         Collection: "collection",
-        Sample: "sample",
-        Relation: "relation",
-        Process: "process",
+        Data: "data",
```
dblenkus, line 127: This reordering should go into a separate commit.

dblenkus (approved):

JureZmrzlikar (approved):
Great effort, all looks good. Let's start using it and see what else is needed!

src/resdk/resources/predictions.py
```diff
@@ -0,0 +1,238 @@
...
+        self.field_id: Optional[int] = None
+
+        #: sample
+        self.sample_id: Optional[int] = None
+        self._sample: Optional[Sample] = None
+        super().__init__(resolwe, **model_data)
+
+    @property
+    @assert_object_exists
+    def modified(self):
```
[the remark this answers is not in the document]
gregorjerse: Since it is actually the modification date from the user perspective.

When new prediction / annotation is created (that overrides the old one) it can be seen as a modification of an old value. Not sure what is the `correct`  way of handling this, I can also just leave `created` field and remove `modified`.

PR #365, opened 2025-01-23
Add inputs field to PredictionField resource

gregorjerse:
as read-only field.


src/resdk/resources/predictions.py
```diff
@@ -62,18 +62,20 @@ class PredictionField(BaseResource):
 
     endpoint = "prediction_field"
 
-    READ_ONLY_FIELDS = BaseResource.READ_ONLY_FIELDS + (
-        "description",
+    READ_ONLY_FIELDS = BaseResource.READ_ONLY_FIELDS
```
dblenkus, line 65: You can omit this line.

gregorjerse: Done.

PR #370, opened 2025-02-06
Add QCTables class

marcellevstek:


src/resdk/tables/qc.py
```diff
@@ -0,0 +1,959 @@
...
+    MACS_PREPEAK = "macs_prepeak_metrics"
+    MACS_POSTPEAK = "macs_postpeak_metrics"
+    PICARD_WGS_METRICS = "picard_wgs_metrics"
+    PICARD_ALIGNMENT_SUMMARY = "picard_alignment_summary"
+    PICARD_DUPLICATION_METRICS = "picard_duplication_metrics"
+    PICARD_INSERT_SIZE_METRICS = "picard_insert_size_metrics"
+    SAMTOOLS_IDXSTATS = "samtools_idxstats"
+    SAMTOOLS_FLAGSTAT = "samtools_flagstat"
+
+    CHIPSEQ_GROUP = {
```
JureZmrzlikar, line 852: The "wiring" in this class is getting complicated more and more. E.g. to add a new table, one must define:
- `GENERAL_STATS_MAP` (un-avoidable)
- GENERAL_STATS constant (nice, but not exactly necessary)
- Add this to various groups (or add it to "independent group, and assign empty dict instead of `*_MAP`")
- Add entry to FILE_MAP
- Write a parser function and possibly modify the `_parse_file`
- ...?

This is getting prohibitively difficult. IMO we should simplify this, or at least keep the data for one data type in one place. Please think about how this could be organized more clearly. ONe example that comes tom mind is the following:

```
# Data types:
    GENERAL_STATS = "general"
    MACS_SUMMARY = "macs_summary"
    MACS_PREPEAK = "macs_prepeak_metrics"
    MACS_POSTPEAK = "macs_postpeak_metrics"
    PICARD_WGS_METRICS = "picard_wgs_metrics"
    PICARD_ALIGNMENT_SUMMARY = "picard_alignment_summary"
    PICARD_DUPLICATION_METRICS = "picard_duplication_metrics"
    PICARD_INSERT_SIZE_METRICS = "picard_insert_size_metrics"
    SAMTOOLS_IDXSTATS = "samtools_idxstats"
    SAMTOOLS_FLAGSTAT = "samtools_flagstat"

    # Data groups
    CHIPSEQ_GROUP = "chipseq"
    WGS_GROUP = "wgs"
    RNASEQ_GROUP = "rna"

    DATA_TYPES = {
        GENERAL_STATS: {
            "file": "multiqc_data/multiqc_general_stats.txt",
            "parser": general_multiqc_parser,
            "column_spec": GENERAL_STATS_MAP,
            "groups": [CHIPSEQ_GROUP, WGS_GROUP, RNASEQ_GROUP],
        },
        MACS_SUMMARY: {
            "file": "multiqc_data/multiqc_macs.txt",
            "parser": general_multiqc_parser,
            "column_spec": MACS_SUMMARY_MAP,
            "groups": [CHIPSEQ_GROUP],
        },
```

This removes the need for `INDEPENDENT_TYPES `, `AGGREGATE_MAP ` and `FILE_PATH_MAP `. It also keeps the infor about a data type a bit more in one place, but I am wondering if you could think of even better solutions...

src/resdk/tables/qc.py
```diff
@@ -0,0 +1,959 @@
...
+            column["slug"]: column["agg_func"]
+            for column in column_names
+            if column["slug"] in df.columns
+        }
+    )
+    series.name = name
+    return series
+
+
+def macs_prepeak_parser(file_object, name, column_names, sample_name):
```
JureZmrzlikar, line 724: A lot of the code repeats in all parser functions. For the sake of simplicity and speed of merging this PR we can leave it as is, but if/when the number of files we wish to parse expands even more we will need to consider how to abstract/ generalize this.

JureZmrzlikar (commented):
A few design-style comments. Otherwise, we are in a very good spot - the gist of work is done, now we need a bit of polishing. Nice work!

src/resdk/tables/qc.py
```diff
@@ -0,0 +1,959 @@
...
+
+import pandas as pd
+
+from resdk.resources import Data
+
+from .base import BaseTables
+
+CHUNK_SIZE = 1000
+
+GENERAL_STATS_MAP = [
```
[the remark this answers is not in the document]
marcellevstek: I have, I think it would be even better to split based on analysis type (fastqc, alignment, downstream QC), etc.

src/resdk/tables/qc.py
marcellevstek: I agree with the comment. Was thinking about having all the information in one dictionary before adding the data groups, but didn't come around implementing it yet. I think that your solution is suitable and don't see a better one currently.

src/resdk/tables/qc.py
```diff
@@ -0,0 +1,1203 @@
...
+    def fetch_group(self, group_name: str) -> pd.DataFrame:
+        """Return a DataFrame with QC data for a specific group."""
+        data_types = [
+            data_type
+            for data_type, data in self.DATA_TYPES.items()
+            if group_name in data["groups"]
+        ]
+
+        if not data_types:
+            raise ValueError(f"Group '{group_name}' not found in available data types.")
```
marcellevstek, line 1184: @JureZmrzlikar should I add a message saying which of the options are available?

marcellevstek: Added

src/resdk/tables/qc.py
```diff
@@ -0,0 +1,1203 @@
...
+        parent_attr = getattr(super(), name, None)
+        if parent_attr is not None:
+            return parent_attr
+        else:
+            if name in self.DATA_TYPES:
+                cached_load_fetch = lru_cache(self._load_fetch)
+                return cached_load_fetch(name)
+            else:
+                raise AttributeError(
+                    f"'{self.__class__.__name__}' object has no attribute '{name}'"
```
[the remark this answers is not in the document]
marcellevstek: Added

src/resdk/tables/qc.py
```diff
@@ -0,0 +1,1221 @@
...
+
+from functools import lru_cache
+
+import pandas as pd
+
+from resdk.resources import Data
+
+from .base import BaseTables
+
+GENERAL_INFO_MAP = [
```
JureZmrzlikar, line 23: These maps take ~700-800 LOC already... For the sake of sanity I am thinking about dumping this "configuration" into a separate file. But on the other hand, it makes sense to have everything in one place... I don'k know, just sharing my thoughts...

src/resdk/tables/qc.py
```diff
@@ -0,0 +1,1221 @@
...
+                raise ValueError(f"Unknown data type: {data_type}")
+
+    def _get_data_uri(self, data: Data, data_type: str) -> str:
+        """Get the file path based on data type."""
+        if data_type in self.DATA_TYPES:
+            return f"{data.id}/{self.DATA_TYPES[data_type]['file']}"
+        raise ValueError(f"Unknown data type: {data_type}")
+
+    @lru_cache()
+    def fetch_group(self, group_name: str) -> pd.DataFrame:
```
JureZmrzlikar, line 1186: Have you considered this API choice with @jkokosar ? I am okay with it, but for consistency, it might still make sense to expose the groups as properties. On the other hand, it is a nice separation that groups are called by a function, while the data types are properties. But the latter distinction is just implementation detail, not a contextual one... I don't know, just speaking out loud.

tests/functional/tables/e2e_tables.py
```diff
@@ -123,3 +123,45 @@ def test_filter(self):
...
+
+    def test_chipseq_qc(self):
+        self.assertEqual(
+            self.qt.fetch_group(group_name="chipseq").loc[
+                139801, "control_prepeak_mapped_percentage"
+            ],
+            0.9018,
+        )
+
+    def test_wgs_qc(self):
```
JureZmrzlikar, line 157: Would be nice to also test one of the property-like data type getters, e.g. `qt.general_alignment` or similar...

JureZmrzlikar (commented):
- Consult @jkokosar on the API choice
- Remove the file `quick_test_qc_tables.py`
- Add the section in the `docs/resdk-tables.rst` and modify the section in `RNATables` about the `.qc` attribute
- Add changelog entry
- Squash commits

Afterwards, we are ready for the final review and merge. 🍾

src/resdk/tables/qc.py
marcellevstek: Agree, will add a separate file.

src/resdk/tables/qc.py
marcellevstek: I have not, @jkokosar please comment on your thoughts about the API choice.

marcellevstek: Not hard, but the code will repeat itself more than it does currently.

docs/resdk-tables.rst
```diff
@@ -5,12 +5,13 @@ ReSDK Tables
 ============
 
 ReSDK tables are helper classes for aggregating collection data in
-tabular format. Currently, we have four flavours:
+tabular format. Currently, we have five flavours:
```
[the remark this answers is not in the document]
marcellevstek: I guess this was forgotten by whoever wrote the code for them.

src/resdk/tables/qc.py
marcellevstek: Fixed, please check if the implementation is appropriate. I don't think this approach is good in terms of readability, but at least it gets rid of some redundancy.

jkokosar (commented):
I've done manual validation by comparing data from MultiQC reports and the output of the QCTables for a random RNA-sequence collection and for "validation run" collection (looking at ChIP-seq data). I've found no further issues beyond those that were already fixed by Marcel. 

API interface is IMO much improved now and enables intuitive interaction with the data. From the content perspective, I consider this done for the current state of pipelines and MultiQC version used. @marcellevstek Please squash the commits. @JureZmrzlikar please give it one final look from the technical perspective.

JureZmrzlikar (commented):
LGTM. Please squash the fixup commit and fix failing linter tests. Than we are good to merge.

src/resdk/tables/qc.py
```diff
@@ -0,0 +1,462 @@
...
+                (sample for sample in self._samples if sample.id == sample_id), None
+            )
+            sample_name = case_sample.name
+            return self.DATA_TYPES[data_type]["parser"](
+                file_object=file_obj,
+                name=sample_id,
+                column_map=self.DATA_TYPES[data_type]["column_map"],
+                sample_name=sample_name,
+            )
+        elif data_type == self.SAMTOOLS_IDXSTATS:
```
JureZmrzlikar, line 409: If you define samtool idxstats parser and qorts genebody parser as:
```
def qorts_genebody_parser(file_object, name, **kwargs):
    # Do nothing with kwargs
```
Than there is no need for special handling of it in this method. You just have the `MACS_PREPEAK` exception and the general case.

JureZmrzlikar (commented):
LGTM. Please squash the commits.

src/resdk/tables/qc.py
marcellevstek: Thanks, forgot about this.

JureZmrzlikar (approved):

PR #373, opened 2025-03-04
Fix ``QCTables`` class to work with the latest MultiQC version

marcellevstek:


src/resdk/tables/qc.py
```diff
@@ -435,7 +441,21 @@ def _parse_file(self, file_obj, sample_id, data_type):
...
-            return f"{data.id}/{self.DATA_TYPES[data_type]['file']}"
+            files = data.files(field_name="report_data")
+            target_fn = self.DATA_TYPES[data_type]["file"]
+
+            if target_fn in files:
+                return f"{data.id}/{target_fn}"
+
+            # Some MultiQC versions returned the same file
+            # with the ``-plot`` suffix
+            file_suffix = Path(target_fn).suffix
```
marcellevstek, line 452: @JureZmrzlikar I don't know if this is the best way to do this, but considering that these are .txt files with only one suffix and there is no .txt in the middle of the filename, it should work as expected.

JureZmrzlikar: It's OK. Code works well, but from aesthetic purpuses the following one would be better IMO:
```
def _get_data_uri(self, data: Data, data_type: str) -> str:
    """Get the file path based on data type."""
    if data_type not in self.DATA_TYPES:
        raise ValueError(f"Unknown data type: {data_type}")

    files = data.files(field_name="report_data")
    target_fn = self.DATA_TYPES[data_type]["file"]

    if target_fn not in files:
        # Some MultiQC versions returned the same file with the ``-plot`` suffix
        suffix = Path(target_fn).suffix
        target_fn2 = target_fn.replace(suffix, f"-plot{suffix}")
        if target_fn2 in files:
            return f"{data.id}/{target_fn2}"

    return f"{data.id}/{target_fn}"
```
It does not have the repeated return statement and the indentation is more flat, which in always beneficial.

JureZmrzlikar (approved):
Looks good, works in the cases that I have tried. Thank you for preparing this!

marcellevstek:
@JureZmrzlikar I amended the PR with your recommendation and added additional code that scales the dataframe column with a specified factor. The problem with the new MultiQC version is that some columns reflecting read counts in general stats table are formatted in millions. Current implementation checks if the source table includes the new column name and scales the values to reflect whole numbers.

JureZmrzlikar:
> @JureZmrzlikar I amended the PR with your recommendation and added additional code that scales the dataframe column with a specified factor. The problem with the new MultiQC version is that some columns reflecting read counts in general stats table are formatted in millions. Current implementation checks if the source table includes the new column name and scales the values to reflect whole numbers.

Looks OK, but please add a comment about this in the method `_filter_and_rename_columns ` for your future self. Or anyone who might be wondering what is the reason for this scaling factor and if it can be removed...

src/resdk/tables/qc.py
```diff
@@ -458,12 +474,38 @@ def __getattr__(self, name):
             return parent_attr
         else:
             if name in self.data_groups:
-                return self._fetch_group(group_name=name)
+                group_data_types = tuple(
```
JureZmrzlikar, line 477: This addition clutters the `__getattr__ ` method, which should be fairly high-level. Please move the new code from `__getattr__ ` to `_fetch_group `.

src/resdk/tables/qc.py
```diff
@@ -458,12 +474,38 @@ def __getattr__(self, name):
...
                 raise AttributeError(
                     f"'{self.__class__.__name__}' object has no attribute '{name}'. "
                     f"Choose one of the following: {list(self.DATA_TYPES.keys()) + self.data_groups}"
                 )
+
+            column_types = {
+                c["slug"]: c["type"] for c in column_mappings if c["slug"] in df.columns
+            }
+
+            df = df.convert_dtypes()
```
JureZmrzlikar, line 503: Why can't you assign the correct dtype and round in `_filter_and_rename_columns` where all the parsing is happening anyway?

marcellevstek: Because` _load_fetch` changes the dtypes.

src/resdk/tables/qc.py
marcellevstek: Ok

src/resdk/tables/qc.py
JureZmrzlikar: OK, bit dtype fix can be solved with `df.convert_dtypes()`. Why are you rounding and setting dtypes also after the line with `df.convert_dtypes()`?

marcellevstek: Because of NaN values, some columns that I want to have them as an integer are casted as floats. I tried doing the ``convert_dtypes()`` in the function you mentioned, but it had zero effect on the output dataframe. Rounding is done to ensure it rounds to nearest integer.

JureZmrzlikar: Ok, I get it. I suggest you overwrite the `_load_fetch()` method in the following way:

```
def _load_fetch(self, data_type):
    df = super()._load_fethc()
   # Do the rounding and casting
```

This should keep `__getattr__` clean(er) and solve the issue in a single place.

src/resdk/tables/qc.py
```diff
@@ -40,18 +41,27 @@
...
+        if isinstance(names, str):
+            names = [names]
+
+        for name in names:
+            if name in df.columns:
+                selected_columns.append(name)
+                rename_map[name] = slug
+                # Some column values are given a specific format (e.g. millions)
+                # and need to be scaled on a common scale
+                df[name] = df[name] * scaling_factor.get(name, 1)
```
JureZmrzlikar, line 62: Once a name is selected further search over names makes no sense. Add `break` to exit the inner for loop.

src/resdk/tables/qc.py
```diff
@@ -448,6 +469,24 @@ def _fetch_group(self, group_name: str) -> pd.DataFrame:
         ]
         data = [self._load_fetch(data_type) for data_type in data_types]
         df = pd.concat(data, axis=1)
+        return df
+
+    @lru_cache()
+    def _load_fetch(self, data_type):
+        df = super()._load_fetch(data_type)
```
JureZmrzlikar, line 476: E2E tests are failing since you are not handling `_load_fetch(self.META)`. Add the following two lines after this one:
```
if data_type == self.META:
    return df
```

JureZmrzlikar (commented):
Looks good, works good. Good job.

Just one minor fix, than we are good to merge.

JureZmrzlikar (approved):

src/resdk/tables/qc_mappings.py
```diff
@@ -3,6 +3,14 @@
 Column mappings with the GENERAL_ prefix are designated for the general summary table in MultiQC.
 In this case, the mappings are not split by file from which the information is extracted
 but based on the type of information.
+
+If ''name'' key includes a list, it means that the column names have changed between MultiQC versions.
```
dblenkus, line 7: Docstrings should be 79 or 99 characters per line. Here, each paragraph exceeds the limit by an increasing margin, making it look worse.

JureZmrzlikar: Thank you for pointing this out. @marcellevstek could you prepare a fix PR?

PR #374, opened 2025-03-10
Add prediction preset

gregorjerse:


src/resdk/resources/predictions.py
```diff
@@ -266,3 +266,92 @@ def __repr__(self):
...
+class PredictionFieldSet:
+    """The set of resources."""
+
+    def __init__(self, parent: BaseResource, field_name: str):
+        """Initialize the set."""
+        self._resources = set()
+        self._field_name = field_name
+        self._parent = parent
+
+    def add(self, *resources: Iterable[Union[BaseResource, int]], patch=True):
```
JureZmrzlikar, line 280: Is there a use-case for `patch=False`?

src/resdk/resources/predictions.py
```diff
@@ -266,3 +266,92 @@ def __repr__(self):
...
+
+    endpoint = "prediction_preset"
+
+    READ_ONLY_FIELDS = BaseResource.READ_ONLY_FIELDS
+
+    UPDATE_PROTECTED_FIELDS = BaseResource.UPDATE_PROTECTED_FIELDS + ("contributor",)
+
+    WRITABLE_FIELDS = BaseResource.WRITABLE_FIELDS + ("name", "fields")
+
+    # fields = SetField()
```
JureZmrzlikar, line 326: Maybe rather `PredictionFieldSet`? Any way, please remove unused comment.

src/resdk/resources/predictions.py
```diff
@@ -266,3 +266,92 @@ def __repr__(self):
...
+
+    @property
+    def fields(self):
+        """Get fields."""
+        return self._fields
+
+    @fields.setter
+    def fields(self, values: Iterable[PredictionField]):
+        """Set fields."""
+        self._fields.clear(patch=False)
```
JureZmrzlikar, line 352: Could this be simplified with just simple `.set()` and 1 request?

src/resdk/resources/predictions.py
JureZmrzlikar: IMO this class is a bit too complicated since it tries to re-implement the `set` object. In essence this is a private class of `PredictionPreset` for managing prediction fields. Based on the current implementation of `PredictionPreset.fields.setter` it should suffice to have only a method that would combine the patch+add method. I don't see any use-case where one would be interested in incremental updates of prediction fields in the `PredictionFieldSet`. One just wishes to create or update the `PredictionPreset`.

I would even go as far to say that we probably do not `PredictionFieldSet` at all. Consider just class `PredictionPreset` and using the builtin `set` object to manage the prediction fields in `PredictionPreset._fields` directly. Can you reuse the patch functionality of BaseResource.save() method instead of defining the patch here?

JureZmrzlikar (commented):
Good start, but IMO we need to agree on API first and unify it with other Resources. Let's talk with all relevant stakeholders.

src/resdk/resources/predictions.py
gregorjerse: Removed.

src/resdk/resources/predictions.py
gregorjerse: Sure, I have added a `set` metod.

src/resdk/resources/predictions.py
gregorjerse: I am not really sure. Exposing a `set` to user directly could lead to undesired side effects (such as changing the preset field set without noticing).

src/resdk/resolwe.py
```diff
@@ -166,7 +173,7 @@ class Resolwe:
 
     session = None
 
-    def __init__(self, username=None, password=None, url=None):
+    def __init__(self, username=None, password=None, cookies=None, url=None):
```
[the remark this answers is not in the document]
gregorjerse: I have removed the commit from the PR.

gregorjerse:
Rewrite in #381

PR #375, opened 2025-03-12
Add fields to ReSDK

gregorjerse:


src/resdk/resources/fields.py
```diff
@@ -0,0 +1,433 @@
...
+        """
+        return value if not self._is_lazy(value) else value()
+
+    def status(self, instance: "BaseResource") -> FieldStatus:
+        """Return the field status."""
+        if not hasattr(instance, self._status_attribute_name):
+            setattr(instance, self._status_attribute_name, FieldStatus.UNSET)
+        return getattr(instance, self._status_attribute_name)
+
+    def changed(self, instance: "BaseResource"):
```
dblenkus, line 182: I don't like that this method requires the parent instance. Is it different from `self._owner`?

dblenkus:
@JureZmrzlikar you don't need to do a detailed technical review, but please run a few common use-cases with this version and confirm that nothing has changed.

src/resdk/resources/fields.py
gregorjerse: The descriptors itself (field instances) are class-level variables, so they have no instances associated with them. The `_owner` is actually a class they are assigned to, not the instance.

So all the methods for working with data require instance as argument.

src/resdk/resources/fields.py
```diff
@@ -0,0 +1,459 @@
...
+            if self._access_type == FieldAccessType.READ_ONLY:
+                raise AttributeError(f"Field {self.public_name} is read only.")
+
+            if self._access_type == FieldAccessType.UPDATE_PROTECTED and created:
+                raise AttributeError(f"Field {self.public_name} is update protected.")
+
+    def _check_many(self, value, instance):
+        """Check that the many attribute is respected.
+
+        :raises AssertionError: the value is not list and many attribute is set.
```
dblenkus, line 118: Fix.

src/resdk/resources/base.py
```diff
@@ -29,25 +38,48 @@ class BaseResource:
...
 
-    READ_ONLY_FIELDS = ("id",)
-    UPDATE_PROTECTED_FIELDS = ()
-    WRITABLE_FIELDS = ()
+    id = IntegerField()
 
     all_permissions = []  # override this in subclass
 
-    def __init__(self, resolwe, **model_data):
+    def __init__(self, resolwe, data_origin_status=DataSource.SERVER, **model_data):
```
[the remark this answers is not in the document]
gregorjerse: Renamed and made user source default.

src/resdk/resources/fields.py
```diff
@@ -0,0 +1,433 @@
...
+        access_type=FieldAccessType.READ_ONLY,
+        required=False,
+        server_field_name=None,
+        assert_exists=False,
+        allow_null=True,
+        initial_loader=None,
+        many=False,
+        accepted_types=None,
+    ):
+        """Initialize the instance.
```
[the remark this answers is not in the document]
gregorjerse: Fixed.

src/resdk/resources/collection.py
```diff
@@ -32,62 +41,32 @@ class BaseCollection(BaseResolweResource):
...
-        "descriptor_dirty",
-        "duplicated",
-    )
-    WRITABLE_FIELDS = BaseResolweResource.WRITABLE_FIELDS + (
-        "description",
-        "descriptor",
-        "descriptor_schema",
-        "settings",
-        "tags",
+    desriptor_dirty = BooleanField()
```
JureZmrzlikar, line 44: Is this still given on the REST API? If yes, is it used anywhere at all?

JureZmrzlikar (commented):
A massive amount of work, kudos. I have few dummy comments, but I would need more time to digest this PR deeply. Will continue with the review later...

tests/functional/docs/e2e_docs.py
```diff
@@ -74,7 +74,7 @@ def upload_reads(self, res):
         reads = res.run(
             slug="upload-fastq-single",
             input={"src": os.path.join(TEST_FILES_DIR, "reads.fastq.gz")},
-            collection=self.collection.id,
+            collection=self.collection,
```
dblenkus, line 77: This will probably break most of the existing ReSDK scripts.

JureZmrzlikar (approved):
I have tried to find an issue with the common tasks, but at this point I can't find one. I know this is not a bulletproof test, but IMO next step forward is to merge and start using it and see where it brakes. Kudos for massive amount of work.

PR #388, opened 2025-04-15
Fix parsing of one-to-one metadata

marcellevstek:


marcellevstek:
@gregorjerse this PR addresses the issue when a one-to-one metadata table object includes samples that are either not present in the collection or don't have a required data object of certain process type (e.g. MultiQC).

Previously the code failed, which posed an issue if some samples imported from GEO failed to process and we wanted to remove them from the collection. When you import through GEO, a one-to-one metadata table is created for each of the imported samples.

Here is some testing code. Try it with this branch versus the master branch to see the difference.

```
import resdk
from resdk.tables import QCTables

res = resdk.Resolwe(url="https://app.genialis.com")
res.login()

collection = res.collection.get("test_metadata")
qt = QCTables(collection=collection)

print(qt.meta)

```

src/resdk/tables/base.py
```diff
@@ -344,14 +336,12 @@ def _get_relations(self) -> pd.DataFrame:
...
-            relations[relation.category] = pd.Series(
-                index=relations.index, dtype="object"
-            )
+            relations[relation.category] = pd.Series(index=relations.index, dtype="object")
 
             for partition in relation.partitions:
                 value = ""
                 if partition["label"] and partition["position"]:
-                    value = f'{partition["label"]} / {partition["position"]}'
+                    value = f"{partition['label']} / {partition['position']}"
```
gregorjerse: Why are these liner changes included in this PR?

mstajdohar:
@gregorjerse please review.

mstajdohar:
@gregorjerse ping. Review should only take a minute, it's just a few lines.

src/resdk/tables/base.py
```diff
@@ -371,6 +371,32 @@ def _get_orange_object(self) -> Data:
         )
 
     def _get_orange_data(self) -> pd.DataFrame:
+        def map_and_filter_samples(
+            df: pd.DataFrame, column_name: str, mapping: dict
```
gregorjerse: Can you be more specific than just `dict`? I believe it is `dict[str, str]`.

src/resdk/tables/base.py
```diff
@@ -371,6 +371,32 @@ def _get_orange_object(self) -> Data:
...
+        ) -> pd.DataFrame:
+            """Map values from a predefined column to a new ``sample_id`` column.
+
+            Samples with no corresponding entry in ``mapping`` are omitted because they are
```
gregorjerse, line 381: I would also mention that the old column is dropped.

src/resdk/tables/base.py
```diff
@@ -402,29 +428,32 @@ def _get_orange_data(self) -> pd.DataFrame:
...
+                df=df, column_name="Sample slug", mapping=mapping
+            )
         elif "mS#Sample slug" in df.columns:
             mapping = {s.slug: s.id for s in self._samples}
-            df["sample_id"] = [mapping[value] for value in df["mS#Sample slug"]]
-            df = df.drop(columns=["mS#Sample slug"])
+            df = map_and_filter_samples(
+                df=df, column_name="mS#Sample slug", mapping=mapping
+            )
         elif "Sample name" in df.columns or "Sample name" in df.columns:
```
gregorjerse: Why is this condition duplicated?

src/resdk/tables/base.py
```diff
@@ -402,29 +428,32 @@ def _get_orange_data(self) -> pd.DataFrame:
             df = df.rename(columns={"mS#Sample ID": "sample_id"})
         elif "Sample slug" in df.columns:
```
gregorjerse: Maybe something like code below? 

```
columns_map = {"Sample_slug": "slug", "mS#Sample slug": "slug", "Sample name": "name", "mS#Sample name": "name"}
for column_name, property in columns_map.items():
    if column_name in df.columns:
        mapping = {getattr(s, property): s.id for s in self._samples}
          if len(mapping) != len(self._samples):
              raise ValueError(
                    "Duplicate sample names. Cannot map orange table data to other metadata"
              )
        df = map_and_filter_samples(
                df=df, column_name=column_name, mapping=mapping
        )

```

gregorjerse (approved):
Code is fine as it is, just some comments.

src/resdk/tables/base.py
marcellevstek: No idea, this was present before and looks like I missed it.

gregorjerse (approved):

gregorjerse (approved):

PR #389, opened 2025-04-16
WIP: Add support for values to query

gregorjerse:
Just a suggestion.

### DIFF
--- a/src/resdk/query.py
+++ b/src/resdk/query.py
@@ -230,8 +230,8 @@ class ResolweQuery:
             resolwe=self.resolwe, **data, initial_data_source=DataSource.SERVER
         )
 
-    def _fetch(self):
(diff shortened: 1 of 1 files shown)


JureZmrzlikar (commented):
I like the idea. A new method on the Query makes it more explicit that you are not receiving the resource objects, but just a dict of requested values. In this regard it seems better API choice than using `fields=[...]` argument in .e.g. filter() method. Let's keep this going with replacing the occurrence of `ResolweQuery.filter(fields=[..])` where appropriate.

mstajdohar:
@gregorjerse, should we continue with a review or close this PR?

mstajdohar:
@dblenkus help @gregorjerse figure out what to do with this PR.

gregorjerse:
No longer necessary.

PR #397, opened 2025-10-01
Fix parsing of MACS prepeak tables

marcellevstek:
in ``QCTables`` for samples with cleaned suffixes


jkokosar:
@gregorjerse please review, and prepare for fix version release when merged.

gregorjerse (approved):

PR #407, opened 2026-05-18
Add a fragment count column

marcellevstek:
in QCTables ``general_fastq`` field output

Previously, raw and trimmed read count columns in QCTables reported the sum across all lanes and mates of FASTQ files per sample. This resulted in inflated number for samples sequenced using paired-end sequencing, since the value was a sum of read counts of both mates. The code now sums across lanes of each mate and then averages the number across both mates. The code assumes that both mates are properly paired. A different aggregation function, for example minimum, could also be used.

REID-3009

### DIFF
--- a/src/resdk/tables/qc.py
+++ b/src/resdk/tables/qc.py
@@ -13,6 +13,7 @@ QCTables
 """
 
 import re
+import warnings
 from contextlib import suppress
 from functools import lru_cache
 from pathlib import Path
@@ -64,8 +65,7 @@ MQC_CLEANED_SFXS = [
 
 def _filter_and_rename_columns(df, column_map):
     """Filter, apply a scaling factor and rename columns based on provided specifications."""
-    selected_columns = []
-    rename_map = {}
+    result = pd.DataFrame(index=df.index)
 
     for col in column_map:
         names = col.get("name", [])
@@ -77,16 +77,12 @@ def _filter_and_rename_columns(df, column_map):
 
         for name in names:
             if name in df.columns:
-                selected_columns.append(name)
-                rename_map[name] = slug
                 # Some column values are given a specific format (e.g. millions)
                 # and need to be scaled on a common scale
-                df[name] = df[name] * scaling_factor.get(name, 1)
+                result[slug] = df[name] * scaling_factor.get(name, 1)
                 break
 
-    df = df[selected_columns].rename(columns=rename_map)
-
-    return df
+    return result
 
 
 def _aggregate_df(df, column_map, prefix=None):
@@ -119,7 +115,95 @@ def _clean_sample_name(
     return sfx_pattern.sub("", filename)
 
 
-def general_multiqc_parser(file_object, name, column_map):
+def _file_name(entry) -> str:
+    """Return the file name from a single Resolwe file field entry."""
+    if isinstance(entry, dict):
+        return entry.get("file", "")
+    return str(entry)
+
+
+def _mate_file_groups(reads_output: dict) -> Optional[list[list[str]]]:
+    """Group cleaned FASTQ file names by read mate from a reads object's output.
+
+    The reads object merges the per-lane input files into a single FASTQ file
+    per mate, exposed on its output regardless of the process that produced it:
+    ``fastq`` (and ``fastq2`` for paired-end libraries). The returned value
+    contains one list of cleaned MultiQC sample names per read mate: two lists
+    for paired-end libraries and a single list for single-end libraries.
+
+    Returns ``None`` if the output does not describe FASTQ files.
+    """
+    if "fastq" in reads_output and "fastq2" in reads_output:
+        mate_keys = ["fastq", "fastq2"]
+    elif "fastq" in reads_output:
+        mate_keys = ["fastq"]
+    else:
+        return None
+
+    return [
+        [
+            _clean_sample_name(Path(_file_name(entry)).name)
+            for entry in (reads_output.get(key) or [])
+        ]
+        for key in mate_keys
+    ]
+
+
+def _sum_mate_read_counts(df, fragment_columns, fastq_names_per_mate, name):
+    """Sum the per-lane read counts within each read mate onto a single row.
+
+    The fragment-count columns hold the FastQC total sequences per FASTQ file.
+    For each read mate (R1/R2), the matching MultiQC rows are summed across
+    sequencing lanes and the total is written to a single row. The downstream
+    mean aggregation (see ``_aggregate_df`` and the ``mean`` agg_func in
+    ``qc_mappings``) then yields the sequenced fragment count: read pairs for
+    paired-end libraries or reads for single-end libraries.
+    """
+    df = df.copy()
+
+    if fastq_names_per_mate is None:
+        warnings.warn(
+            "Could not resolve the FASTQ file names from the reads object "
+            "that was used as MultiQC input for "
+            f"sample {name}; fragment counts are left empty.",
+            UserWarning,
+        )
+        df.loc[:, fragment_columns] = pd.NA
+        return df
+
+    # MultiQC renders FastQC row labels in general stats table as "<sample_name> | <fastq_name>";
+    # the part after the "|" can be matched to reads object file name outputs. Only labels with
+    # exactly one "|" carry a FASTQ name; any other label does not and is ignored.
+    mqc_fastqc_row_labels = df.index.map(
+        lambda x: str(x).split("|")[1].strip() if str(x).count("|") == 1 else pd.NA
+    )
+    for fragment_col in fragment_columns:
+        per_mate_sums = {}
+
+        for mate_fastq_names in fastq_names_per_mate:
+            matching_rows = (
+                mqc_fastqc_row_labels.isin(mate_fastq_names) & df[fragment_col].notna()
+            )
+
+            if not matching_rows.any():
+                continue
+
+            # The first MultiQC row label is used as the index
+            placeholder_index = df.index[matching_rows][0]
+
+            # Sum across lanes values per mate
+            per_mate_sums[placeholder_index] = df.loc[matching_rows, fragment_col].sum()
+
+        # Wipe original values (total read counts - see qc_mappings)
+        df[fragment_col] = pd.NA
+
+        for placeholder_index, per_mate_sum in per_mate_sums.items():
+            df.loc[placeholder_index, fragment_col] = per_mate_sum
+
+    return df
+
+
+def general_multiqc_parser(file_object, name, column_map, fastq_names_per_mate=None):
     """General parser for MultiQC files.
 
     If the column_map argument is not specified the function will retain the original column names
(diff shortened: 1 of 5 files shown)


mstajdohar:
What’s the latest on this PR? Should we close it? Otherwise, what’s blocking a merge?

gregorjerse (commented):
**Adversarial review — High/Medium findings** (Low-severity items and nits omitted). Per-line comments below.

Two items not tied to a specific changed line:

- **[Medium] No tests / no fixture.** The new branching (`_is_paired_end`, the per-mate `_1`/`_2` block) is unexercised — the diff touches no `tests/`, and there is no `multiqc_general_stats` fixture in the repo, so the central `_1`/`_2` naming assumption is unvalidated. A multi-lane paired-end fixture asserting the exact fragment count is exactly what would catch the naming issue flagged inline.
- **[Medium] CHANGELOG.** `docs/CHANGELOG.rst` has no `Unreleased` entry for the two new user-visible columns and the changed paired-end read-count semantics.

Credit: the `_filter_and_rename_columns` rewrite to a fresh `result` DataFrame is correct and now load-bearing — with two slugs (`total_read_count_*` and `estimated_fragment_count_*`) sharing one source column, the old in-place `df[name] *= ...` + `rename_map` approach would double-scale and collide column names. Please don't revert it to in-place mutation.

src/resdk/tables/qc.py
```diff
@@ -119,7 +114,22 @@ def _clean_sample_name(
     return sfx_pattern.sub("", filename)
 
 
-def general_multiqc_parser(file_object, name, column_map):
+def _is_paired_end(sample) -> Optional[bool]:
+    reads = list(sample.data.filter(type="data:reads:fastq:"))
```
gregorjerse: **[High] Blocking, synchronous network call inside the asyncio download loop.** `sample.data.filter(...)` materialized with `list()` runs `ResolweQuery._fetch → self.api.get(...)` — a blocking slumber/`requests` GET. It executes inside `general_multiqc_parser → _parse_file`, which is called from the `async def _download_file` coroutine gathered by `asyncio.gather` (`base.py`), with no executor / `to_thread`. So it stalls the event loop and serializes the very downloads the chunked async path exists to parallelize, adding a round-trip per parsed file. It is also uncached (`.filter()` clones with an empty cache; no memoization on `_is_paired_end`).

Suggest resolving library strategy once per sample *off* the event loop (e.g. precompute a `{sample_id: Optional[bool]}` map) and passing the boolean into the parser.

src/resdk/tables/qc.py
```diff
@@ -119,7 +114,22 @@ def _clean_sample_name(
...
+    origin_read_obj = reads[-1] if reads else None
```
gregorjerse: **[Medium] `reads[-1]` is non-deterministic and `origin_read_obj` is misleading.** No `ordering=` is set, so the "last" element depends on undefined server ordering, yet the variable is named `origin_read_obj` while taking the *last* item. The codebase idiom (`shortcuts/sample.py` `get_reads`) pins `ordering='-id'` and takes `[0]`. Combined with the strict `endswith` just below, *which* reads object is picked actually matters. Pin an explicit order (or tie selection to the reads upstream of the MultiQC `Data`) and rename to match what is selected.

src/resdk/tables/qc.py
```diff
@@ -119,7 +114,22 @@ def _clean_sample_name(
...
+    if origin_read_obj is None:
```
gregorjerse: **[Medium] `None` (unknown strategy) is silently treated as single-end.** The call site is `if _is_paired_end(sample):`, so a `None` return falls into the single-end branch, which sums across *both* mates — re-introducing exactly the 2× inflation this PR fixes, with no warning, for any genuinely paired-end sample whose reads object can't be resolved here. Handle the three states explicitly (e.g. warn or skip the adjustment on `None`), consistent with the `UserWarning` pattern already used in `base.py`.

src/resdk/tables/qc.py
```diff
@@ -119,7 +114,22 @@ def _clean_sample_name(
...
+        return None
+    if origin_read_obj._process._type.endswith("paired:"):
```
gregorjerse: **[High] Strict `endswith('paired:')` / `('single:')` raises on processed reads, and the failure takes down unrelated tables.** Upload types like `data:reads:fastq:paired:` match, but a *trimmed* object — e.g. `data:reads:fastq:single:cutadapt:` (this exact shape is in `tests/unit/test_sample.py`) — ends in `cutadapt:` and hits the `else: raise ValueError`. Since `reads[-1]` often resolves to the higher-id (trimmed) object, this is plausible rather than exotic. And because `_is_paired_end` runs for ~15 data types via `general_multiqc_parser` while `asyncio.gather` is called without `return_exceptions=True`, one such sample aborts parsing of picard / flagstat / bowtie2 / etc. for the whole chunk — coupling that did not exist before this PR.

Match the library token as a path segment instead of a strict suffix (e.g. `'paired' in type.split(':')`), and make the unknown case non-fatal.

src/resdk/tables/qc.py
```diff
@@ -138,6 +148,34 @@ def general_multiqc_parser(file_object, name, column_map):
...
+    # and insert the resulting value into the first column for the corresponding mate.
+    # This is is done in order to report the actual read pairs for samples
+    # sequenced in paired-end mode, while preserving the sum across lanes.
+    fragment_metrics = [
+        "estimated_fragment_count_raw",
+        "estimated_fragment_count_trimmed",
+    ]
+    fragment_columns = df.columns.intersection(fragment_metrics)
+
+    if _is_paired_end(sample):
```
gregorjerse: **[Medium] `_is_paired_end` is called unconditionally, even when there are no fragment columns.** `fragment_columns` is computed just above but not used to gate this call. For the ~14 non-FASTQ data types routed through this parser, `fragment_columns` is always empty, so the per-mate loops are a no-op — yet the blocking reads query (and the `ValueError` risk above) is still paid for every sample. A single guard removes the waste *and* confines the crash surface to the one table that needs it:

```python
if not fragment_columns.empty:
    paired = _is_paired_end(sample)
    ...  # the whole paired/single block, only here
```

src/resdk/tables/qc.py
```diff
@@ -138,6 +148,34 @@ def general_multiqc_parser(file_object, name, column_map):
...
+        for col in fragment_columns:
+            for s in ["_1", "_2"]:
+                mate_mask = df.index.str.endswith(s) & df[col].notna()
```
gregorjerse: **[High] The `_1`/`_2` mate-detection assumption is unvalidated and fails silently for common naming; it also crashes on non-string indexes.**

(1) `endswith('_1')` / `('_2')` matches SRA-style `sample_1` / `sample_2` but **not** Illumina `sample_R1_001` / `sample_R2_001` or `_R1` / `_R2` — `'sample_R1'.endswith('_1')` is `False`, and `MQC_CLEANED_SFXS` strips `.fastq` / `.gz` but not the `_R1_001` tail. When neither matches, both masks are empty, `continue` fires, and `_aggregate_df` averages all `2L` lane×mate rows → the reported value is `sum / 2L` instead of `sum / 2`, i.e. **`1/L` of correct** (under-reporting that scales with lane count). A single-lane sample returns the right number *by coincidence*, so a naive fixture won't catch it. Please confirm what suffix the Genialis MultiQC config actually emits; safest is to derive mate identity from the fastq `Data`'s R1/R2 files, or at minimum match `_(?:R)?[12](?:_\d+)?$` and warn/raise when a paired sample matches neither mate instead of silently averaging.

(2) `df.index.str.endswith(...)` raises `AttributeError: Can only use .str accessor with string values` when sample names are purely numeric (`read_csv(index_col=0).convert_dtypes()` yields an integer index). The single-end branch isn't affected, so the failure is asymmetric. Use `df.index.astype(str).str.endswith(s)`.

src/resdk/tables/qc.py
```diff
@@ -138,6 +148,34 @@ def general_multiqc_parser(file_object, name, column_map):
...
+                if not mate_mask.any():
+                    continue
+                total = df.loc[mate_mask, col].sum(min_count=1)
```
gregorjerse: **[Medium] "Mean of the two mate totals" is an estimator that degrades silently.** When only one mate's FastQC row is present, the other mate's mask is empty and `mean` over `[mate1_total, NA…]` returns the single-mate count, not a pair count. For the *trimmed* column, mates are trimmed independently and can legitimately differ, so the mean is a fractional value that is neither mate's count nor a true pair count (and disagrees by design with `total_read_count_*`, which still sums everything). Consider defining fragment count as R1's lane-summed total (== pair count by construction, immune to mate-2 asymmetry), or document the estimator and warn when `|m1 − m2|` is large or a mate is missing.

src/resdk/tables/qc.py
marcellevstek: The ordering is determined by the backend, so this is not entirely applicable.

src/resdk/tables/qc.py
marcellevstek: This is an insightful comment, as the code was written with assumptions of the user running the General RNA-seq pipeline. Will modify to make it more general.

src/resdk/tables/qc.py
marcellevstek: Will test further

mstajdohar:
@marcellevstek Would you address the feedback from @gregorjerse?

marcellevstek:
@mstajdohar I don't think there is a straightforward way to solve this, since we have to rely on sample names that can be arbitrary. I think we can close this PR, but keep in mind the issues that it was planned to solve.

mstajdohar:
Discuss with @jkokosar on how to proceed.

src/resdk/tables/qc.py
marcellevstek: I understand the concern about the trimmed fragment counts, but the column name itself states that this is an estimated fragment count. I don't think there is a better way of presenting the data, but I am open to new ideas. The values of both mates are usually really close, so this should not be an issue.

src/resdk/tables/qc.py
marcellevstek: Fixed in the latest commit. This was not done optimally and I now match using the raw reads output file names. There are still a few assumptions in the code based on how MultiQC presents the data in the table that is parsed. But looking at a few different cases (also captured in tests) it parses the data correctly.

src/resdk/tables/qc.py
marcellevstek: The description of the PR also states: The code assumes that both mates are properly paired.

src/resdk/tables/qc.py
```diff
@@ -451,29 +549,67 @@ def __init__(
...
+        input_ids = data_obj.input.get("data") or []
+        if not input_ids:
+            return None
+
+        # Use the oldest reads object, i.e. the uploaded reads rather than a
+        # downstream (e.g. trimmed) object. Its merged per-mate FASTQ files are
+        # read from the output, which is consistent across producing processes.
+        reads = next(
+            iter(
+                self.resolwe.data.filter(
```
gregorjerse: **Performance: blocking HTTP request inside the asyncio event loop**

`_mate_fastq_groups` issues a synchronous `self.resolwe.data.filter(...)` request. It is reached via `_parse_file` → `_download_file`, which runs under `asyncio.gather` in `BaseTables._download_data`. A blocking `requests` call inside a running coroutine stalls the event loop, so:

- the per-sample reads-object lookups effectively serialize, and
- they block the concurrent file downloads the async path was built to parallelize.

Net effect is one extra sequential API round-trip per sample injected into `general_fastq` download — a noticeable regression on large collections.

Suggestion: prefetch the reads objects in a single batched query before the async download starts (you already have every sample's MultiQC `input["data"]` ids), build a `sample_id -> mate groups` map once, and have `_mate_fastq_groups` read from that map instead of hitting the API from inside the coroutine.

gregorjerse (approved):

src/resdk/tables/qc.py
```diff
@@ -451,29 +551,91 @@ def __init__(
             set(group for data in self.DATA_TYPES.values() for group in data["groups"])
         )
 
+    @lru_cache()
+    def _mate_fastq_groups(self) -> Dict[int, Optional[List[List[str]]]]:
```
gregorjerse: Please, replace `Dict` with `dict`.

PR #409, opened 2026-08-04
Skip downloading of intact files and streamline the download functionality

marcellevstek:


src/resdk/resources/data.py
```diff
@@ -290,17 +300,71 @@ def download(
...
 
+        dir_names = self._files_dirs("dir", file_name, field_name)
+
+        if custom_asset_name:
+            # Files of a downloaded directory are listed in file_names as well,
+            # so a matched directory counts as a single asset.
+            matched_assets = dir_names if dir_names else file_names
+            if len(matched_assets) != 1:
+                raise ValueError(
+                    f"custom_asset_name can only be given when a single asset is "
```
gregorjerse: Please change to something nicer. Like

f"A 'custom_asset_name' can only be applied when downloading exactly one asset. Found {matched_assets} matching assets instead."

src/resdk/resources/data.py
```diff
@@ -290,17 +300,71 @@ def download(
...
+            self.resolwe._download_files(
+                files=files,
+                download_dir=download_dir,
+                show_progress=show_progress,
+                custom_file_name=custom_asset_name,
+            )
+            return [custom_asset_name]
 
-        return asset_names
+        if download_dir is None:
```
gregorjerse: If you preffer one-liners this one could be

`download_dir = download_dir or os.getcwd()`

